PR #43, opened 2025-01-30
added markdown conversion

cburgard:
Since there was not much movement in the overleaf in the last weeks, I figured it might be a good opportunity to perform the conversion of the overleaf document (tex) to markdown.

For this purpose, I used pandoc.

In order to nicely view and render the output, again, pandoc can be used.
I've added commands to do the conversion to a reasonably-well looking version of pdf and html to a Makefile which I added.

If you want to take a look at the result, just check out the branch and type `make`. If you have pandoc installed, it should give you static HTML and PDF renderings of the HS3 document that are reasonably-well formatted.

There are still a lot more things to be done, such as
- adding a static HTML generator for a dynamic webpage
- sanitizing the math
- improving the bibliography
- adding a CI to statically deploy pages

But I figured this is a good moment to ask people for feedback.
I'd be especially curious for input from @kratsg, since (if I remember correctly) he was the one to suggest markdown.


kratsg:
This is great first start. What I'm thinking about right now is what the "core" data should look like. That is, imagine that we write it all in markdown/rST and use that to generate all other outputs (LaTeX, HTML, ePub, PDF, etc) -- or if we start even lower-level with something like JSON or YAML data files -- using that to generate the markdown which then generates the rest.

For the LaTeX to HTML, there does exist https://github.com/brucemiller/LaTeXML which arXiv is using, and it's really nice and gives us something that can just be hosted directly on a single page. I was wondering if it would make sense to support this documentation via something like readthedocs using either sphinx or mkdocs for "generating", allowing us to produce a uniform structure with search capability, etc..?

cburgard:
> This is great first start. What I'm thinking about right now is what the "core" data should look like. That is, imagine that we write it all in markdown/rST and use that to generate all other outputs (LaTeX, HTML, ePub, PDF, etc) -- or if we start even lower-level with something like JSON or YAML data files -- using that to generate the markdown which then generates the rest.
> 
> For the LaTeX to HTML, there does exist https://github.com/brucemiller/LaTeXML which arXiv is using, and it's really nice and gives us something that can just be hosted directly on a single page. I was wondering if it would make sense to support this documentation via something like readthedocs using either sphinx or mkdocs for "generating", allowing us to produce a uniform structure with search capability, etc..?

Thanks for the quick reply.

In case this was not clear: The content of this MR is intended to become the new "master document", that is, the overleaf will be archived and this version (markdown) will become the source document.

That way, all the things you advocate for here should be possible quite easily.

I experimented already with static webhosting (jekyll) and conversion to epub, but was not very happy with the results, hence I stuck with the html and tex/pdf outputs for my Makefile.

But what I gather from this is that you are in support of making this (markdown) version the new master document?

kratsg:
I am in favor :) I think we can use pure markdown and figure out how to do more with it/formatting/exporting etc.

PR #78, opened 2026-02-23
GenNormal distribution

agbuckley:
Adds a generalised normal distribution. And a rather important bugfix (or two) for the Gaussian PDF formula!

### DIFF
--- a/docs/chapters/2.1.1_fundamental_distributions.md
+++ b/docs/chapters/2.1.1_fundamental_distributions.md
@@ -123,7 +123,7 @@ $$
 The Gaussian/Normal distribution is defined as 
 
 $$
-\begin{aligned} \text{GaussianPdf}(x, \mu, \sigma) = \frac{1}{\ensuremath{\mathcal{M}}} \exp\left(-\frac{(x-\mu)^2}{\sigma^2}\right) \end{aligned}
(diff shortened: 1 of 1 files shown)


cburgard:
Hi Andy,

Thanks a lot for adding this. I am happy with your proposed changes, I have just one request: Perhaps it would be better to spell out `generalized_normal_dist`. Here's my rationale for this suggestion:

- it is more in line with the `multivariate_normal_dist` this way
- "gen" as a prefix could be misunderstood to mean "generative" or "generic"
- I think we generally spell out distribution names fully. Looking through the list of distributions, the only one I could find that is arguably an abbreviation is the `lognormal_dist`, where I would argue that `lognormal` is the usual colloquial name, rather than spelling it out further. So unless the same is the case for the generalized normal (which I don't know), I think that is not a good reference.

Please let me know what you think.

agbuckley:
Hi @cburgard , no problem at all with that -- I'm usually a fan of "compact but not ambiguous" as a naming strategy, but perhaps "generative" will become a point of confusion, and for sure the principles of the naming scheme, whatever they are, should be the same through the spec: you are the guardian of that! So no objection from me to any naming changes, it's just the concept I have an opinion on ;-)

cburgard:
Ok thanks for the fast reply, I've performed the renaming and made sure it uses the "-ized" spelling (same as everywhere else). If you're happy with this, I'll merge.

PR #79, opened 2026-02-24
Fully parametrised data uncertainties

agbuckley:
A draft proposal for an error-source/templated/modifier-based data uncertainty specification. To be discussed!

### DIFF
--- a/docs/chapters/2.1.1_fundamental_distributions.md
+++ b/docs/chapters/2.1.1_fundamental_distributions.md
@@ -31,7 +31,7 @@ and describes the ARGUS background shape.
 
 #### Continued Poisson distribution
 
-The of a continued Poisson distribution of the variable $x$ is defined as 
+The continued Poisson distribution of the variable $x$ is defined as 
 
 $$
 \begin{aligned} \text{ContinuedPoissonPdf}(x, \lambda) = \frac{1}{\ensuremath{\mathcal{M}}} \exp\left(x \cdot \ln \lambda - \lambda - \ln \Gamma(x+1)\right), \end{aligned} 
@@ -133,6 +133,7 @@ $$
 -   `x`: name of the variable $x$ 
 -   `mean`: value or name of the parameter used as mean value $\mu$ 
 -   `sigma`: value or name of the parameter encoding the standard     deviation $\sigma$. 
+
 #### Log-Normal distribution
 
 The log-normal distribution is defined as 
@@ -187,7 +188,7 @@ $$
 This section contains multivariate fundamental distributions. They may
 refer to functions, parameters and more than one variable.
 
-#### Barlow-Besston-Lite Constraint distribution
+#### Barlow-Beeston-Lite Constraint distribution
 
 This distribution represents a product of Poisson distributions
 defining the statistical uncertainties of the histogram templates
@@ -201,7 +202,7 @@ $$
 
 -   `name`: custom unique string 
 -   `type`: `barlow_beeston_lite_poisson_constraint_dist` 
--   `x`: name of the variable $x$ 
+-   `x`: array of names of the variables $x_i$. This also includes mixed arrays of values and names. 
 -   `expected`: array of central values $\tau_i$ 
 
 #### Multivariate normal distribution
--- a/docs/chapters/2.3_data.md
+++ b/docs/chapters/2.3_data.md
@@ -16,9 +16,11 @@ While data, in most settings, has no uncertainty attached to it, this standard *
 -   generated data obtained from importance sampling, including a     (potentially Gaussian) uncertainty from the frequency weights 
 -   unfolded data, resulting from arbitrarily complex transformation     functions involving statistical models folding some degree of     uncertainty into the data points themselves 
 
-While it should always be preferred to publish "raw" data, allowing to include pre-processed data with corresponding uncertainties expands the possible applications considerably. 
+While it should always be preferred to publish "raw" data, e.g. along with the statistical model used for its unfolding to inferred particle-level observables, allowing use of pre-processed data with corresponding uncertainties expands the possible applications considerably.
+The uncertainties encoded in this way on the data will normally need to be mapped into the statistical model with appropriate nuisance parameters, which can be named in declaration of the data-uncertainties using the same form as for distributions/constraints.
 
-### Point Data 
+
+### Point data 
 Point data describes a measurement of a single number, with a possible uncertainty (error). 
 
 -   `name`: custom string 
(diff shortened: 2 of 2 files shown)


cburgard:
HI Andy,

Thanks a lot for this proposal. This part of our spec has been written without a clear usage example in mind, and you picking this up and working out details that are actually useful in a concrete case is highly appreciated.

I have a few thoughts about your proposal that we can discuss.

First of all, I like that you specifically allow for different types of uncertainties such as "gaussian_uncertainty" and your new "modifier_uncertainty". This opens the pathway to in the future also support other, new types that we have not thought of yet. 

The "modifier_uncertainty" you propose makes sense to me in order to decompose several types of uncertainties. However, what is not clear to me is what the implicit assumptions here are:

- is the idea that each modifier has a parameter that models this uncertainty? is it allowed for these parameters to be identical with parameters of the model itself, so that uncertainties in data and model can cancel in nontrivail ways?
- is it allowed for these parameters to have constraint terms? how are they referenced?
- right now, your proposal heavily references a specific distribution type, relying on the same type of syntax. in the past, we have taken such cases and factorized out the common component of the spec to a section in the appendix (see "generic_function" and "generic_distribution" for examples). Maybe it would be better to do this here as well in order to make sure that future changes to one of the two will automatically be in sync with the other

As both gaussian and histfactory are distributions in the standard, I am also thinking about whether it would not make sense to factorize out the description of the uncertainty to an actual explicit distribution, perhaps in the following style:

```
"uncertainty": "my_multivariate_gauss"
```
Where "my_multivariate_gauss" would then be a gaussian distribution listed under "distributions". This would immediately open up the possibilty to use any distribution type as an uncertainty parametrization, rather than basically duplicating parts of the standard for distributions here in the data section. But perhaps this amount of generality is also not ideal, as this might make it difficult for framework developers to implement this correctly, as this might still prompt duplicate implementations on the framework side to implement the data-version and the model-version of each distribution.

I"d be curious what your thoughts on this are. I'm happy to go along with what you propose here, as it clearly fills a real need, but I'm also thinking into the future which other functionality we might want to have and how to make sure all the pieces are lined out in a way to provide us with this flexibility in the future without introducing unneccessary complexity or duplication. Please let me know what you think.

Also, I'd be very interested in reflections from other people on this topic, especially from @oschulz as the current version of this section resulted from a discussion with him.

agbuckley:
Hi @cburgard , thanks for the thoughts, and sorry to take a few days to get time to read and respond!

To quickly clear up some of the the questions: yes, each uncertainty modifier/template would have an associated parameter in the statistical model: the parameter would naively have the name given by the name given to that error source in e.g. HEPData. This will often match the reconstruction nuisance-parameter names, i.e. for various eras of jet or lepton reconstruction, flavour tagging, etc. -- as the "data" for unfolded measurements is already the output of an inference, "data uncertainties" here often have the same origins as "model uncertainties" where the data is reco-level event counts. (For this reason, if you wanted to be able to combine measurements and searches you would want to preserve those names, rather than e.g. prepend `DATA_` to the data-uncertainty parameters to ensure their uniqueness.

I'm not sure where best to put these parameters: obviously the effect is the same in terms of test-statistic value between the model and data if you e.g. add/subtract a shift to the data or subtract/add a shift to the model, but really what is moving is the data and I think it would be confusing to invert that. In the end, the model will contain terms for data and prediction, and nuisance parameters will be able to modify the values of both (but _usually_ POIs only for the prediction; the potential counter-example is something like a top-quark EFT fit where the expected top-reconstruction performance itself depends on the presence of non-zero EFT contributions and so you would want those physics-model parameters present as unfolding nuisances in the inference of differential cross-section bin-values -- we saw this for https://arxiv.org/abs/2303.03427)

These terms will be constrained: the variations are usually reported as +-1 sigma deviations of the bin values for each error source, where the deviations are generally asymmetric but the parameter is typically assumed to be constrained by a unit Gaussian. This isn't _necessarily_ the model, though: I think it should be possible to change the constraint, but perhaps make the Gaussian way a default that doesn't require an explicit constraint specification? I can imagine it potentially being useful to play with the constraints, and also to be able to provide more detailed control over the templates, e.g. for more than two nuisance values, and at custom nuisance-parameter values; is there currently a way to do that for distributions, and to control the interpolation method between HF modifiers?

I agree, definitely good if duplication can be avoided, and synchronisation designed to be automatic. This is a (common) use-case that blurs the distinction between model and data, as both can be functions of fit parameters... I don't know HS3 well enough to have a very clear view of the best / most-HS3-y way, but happy to discuss and iterate: this was definitely a first-draft proposal to start discussion! All the best.

cburgard:
Hi @agbuckley ,

my preliminary idea would be something like this, but please feel free to tell me this is nonsense:

```
"data" : [ {
  "name": "my_data" ,
  "type": "binned",
  ...
  "uncertainty" : { 
    "type" : "distribution",
    "distribution" : "my_stack"
  }
}],
"distributions" : [ {
  "name" : "my_stack",
  "type" : "histfactory_dist",
  "samples: [
     { "name":"combined", "data" : ... "modifiers" : [
    ... here goes the modifiers ...
   ]}]
}]
```

The problem here is of course that a lot of info is not very useful. Like, what does it mean that the distribution used to model the data uncertainty has a mean of its own - should this always be 0, or should this be always identical for the data yield for consistency reasons?
I'm not sure if this is a sane approach, but I am tempted by the possibility to just "refer to" an existing distribution. That way, we could make sure that the uncertainties allowed for data and model are always kept in sync.

What do you think?

agbuckley:
Thanks @cburgard , and sorry it's taken time to get back to this -- I don't have the familiarity with the HS3 spec needed to draw immediate conclusions! My feeling is that it would be best to use the syntax/structure for the distributions, but locate it _in_ the data block.

That is mainly because I worry about the coherence of a format that distributes naturally connected things -- a set of values and their uncertainties -- to very different places in the file. There is a benefit to being able to see those parts separated only by a line or two, and to therefore see that the structure agrees, that the errors are of the appropriate scale, etc. Having to view the file with two different windows to see the different bits of the dataset would be a red flag! Of course, being delocalised is no problem for a parser, but being amenable to human inspection is desirable.

Localising the histfactory/distribution/template-based uncertainties near the data would also be consistent with defining the simpler Gaussian error-list or covariance/correlation-matrix elements in the data block: I don't think you   are proposing to delocalise those, too? Splitting the general variation-template parametrisation out to the distributions block but keeping their simplified forms in the data seems counterintuitive.

If anything, this discussion maybe flags that at the moment the distributions are (I think) not an abstract set of definitions to be used by both data and models, but are designed around the models. So there's an asymmetry: model data and uncertainties for histfactory type sit in the same place naturally, but the data would be awkwardly split, mostly because of the history of how the models and distributions were designed. I don't think that's desirable, but hopefully using the same structure in two different places would mean that code duplication can be avoided, by initialising the object from the histfactory config dict in the same way regardless of where that dict was located in the file.

Thoughts?

cburgard:
This PR is out of date and has been superseded by developments in the associated discussion thread https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/discussions/88

PR #83, opened 2026-03-04
Conversion to typst

oschulz:
For now just an experiment to see if Typst would give us nicer math and PDF generation.

The main document `hs3.typ` compiles and looks correct, but no CI resp. automatic generation is set up yet. The Markdown files are sill present and untouched, for direct comparison.

Conversion of Markdown to Typst via Claude 4.5 Opus.

### DIFF
--- a/docs/chapters/1_introduction.typ
+++ b/docs/chapters/1_introduction.typ
@@ -1,0 +1,96 @@
+#import "/docs/hs3-defs.typ": sc
+// Chapter 1: Introduction
+
+= Introduction <sec:introduction>
+
+_This section is non-normative._
+
+It is widely agreed upon that the publication of likelihood models from high energy physics experiments is imperative for the preservation and public access to these results. Detailed arguments have been made elsewhere already @Cranmer_2022. This document sets out to provide a standardized format to publish and archive statistical models of any size and across a wide range of mathematical formalisms in a way that is readable by both humans and machines.
+With the introduction of `pyhf` @pyhf @pyhf-joss, a `JSON` format for likelihood serialization has been put forward. However, an interoperable format that encompasses likelihoods with a scope beyond stacks of binned histograms was lacking. With the release of `ROOT` 6.26/00 @root and the experimental `RooJSONFactoryWSTool` therein, this gap has now been filled.
+This document sets out to document the syntax and features of the Statistics Serialization Standard (HS#super[3]) for likelihoods and statistical models in general, as to be adopted by any HS#super[3]-compatible statistics framework. The examples in this document employ the `JSON` notation, but are intended to encompass also representations as `YAML` or `TOML`.
+
+== How to use this document
+
+Developers of statistical toolkits are invited to specifically refer to versions of this document relating to the support of this standard in their respective implementations.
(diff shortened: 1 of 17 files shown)


kratsg:
> For now just an experiment to see if Typst would give us nicer math and PDF generation.
> 
> The main document `hs3.typ` compiles and looks correct, but no CI resp. automatic generation is set up yet. The Markdown files are sill present and untouched, for direct comparison.

we'd still need to allow cross-reference to this documentation through a inventory or tags file generated but it looks like typst doesn't support that.

cburgard:
We can for sure use typst to generate pdfs, but I'm more wondering if we want to use at as the original source.
To me, keeping the source to be markdown makes a lot of sense.
The PDF generation at the moment happens by dynamically generatig tex using pandoc and then going to pdf from there. Maybe we can do something similar, but with typst as an intermediate format?

kratsg:
Is there an example of the comparison between the two generated outputs to refer to?

oschulz:
> Is there an example of the comparison between the two generated outputs to refer to?

Not yet.

oschulz:
This is really just an quick experiment for now. As @cburgard points out, there are clear advantages to keeping the source content in Markdown form. One can then, for example, easily integrate parts of it into software package docs.

Generating Typst via pandoc primarily for PDF generation would be a possibility, and would be much more lightweight than requiring a LaTeX installation.

For now I just wanted to see how easily and faithfully one can move from Markdown to Typst via LLM, and what the results look like. A comparison with a (much faster and cheaper) pandoc generation would be interesting, for example - I expect pandoc will do very well, in fact.

kratsg:
> Generating Typst via pandoc primarily for PDF generation would be a possibility, and would be much more lightweight than requiring a LaTeX installation.

we always have MathJax which performs quite well -- if the problem is just ending up having a PDF generated, that's a separate issue we can deal with right?

oschulz:
> we always have MathJax which performs quite well -- if the problem is just ending up having a PDF generated, that's a separate issue we can deal with right?

Sure, though I'm curious if Tyst can produce nicer math than MathJax. Like I said, I'm just playing around here.

kratsg:
My understanding is that typst is a subset of latex, and when I had tried it a while back, ended up going back to latex since it didn't offer all the control that was needed. If the issue is more about getting latex set up, one can use [pixi](https://pixi.prefix.dev/latest/) and deal with using latex.

It would be good to get a comparison and understand the pros/cons. I tend to err on using a tool that more people (who contribute to this) are familiar with.

oschulz:
> My understanding is that typst is a subset of latex

Not really, it's a new typesetting language that offers clean versioned package management, a lightweight fast engine, and partial recompilation. As a language, Typst is Turing-complete, like LaTeX, but the design of the language allows for faster execution. But yes, there are things where LaTeX still has an edge in functionality still, here or there, from what I know - not massively anymore, I think, Typst is maturing quickly. When it comes to PDF or HTML+math generation from Markdown that shouldn't matter anyway, as Markdown's feature set is so much smaller.

oschulz:
I think we'll stay with Markdown.

PR #84, opened 2026-03-04
Add function outputs and reference syntax.

oschulz:
This PR unifies the role of the name for functions and distributions. It also provides the necessary basis for multiple-output functions and hierarchical models.

In the future, when adding support for vector-values, we can then also allow `"function_name.output_name[index]"` and `"distribution_name.variate_name[index]"`.

### DIFF
--- a/docs/chapters/2.1_distributions.md
+++ b/docs/chapters/2.1_distributions.md
@@ -15,19 +15,25 @@ Each distribution [must]{.smallcaps} have the components `type`, denoting the ki
 	{ 
 		"name":"gauss1", 
 		"type":"gaussian_dist", 
-		"mean":1.0, 
+		"mean":"sum_1", 
 		"sigma":"param_sigma", 
-		"x":"param_x" 
+		"x":"data_x" 
 	}, 
 	{ 
 		"name":"exp1", 
 		"type":"exponential_dist", 
 		"c":-2, 
-		"x":"data_x" 
+		"x":"param_x" 
 	}, 
 	... 
 ] 
 ```
+
+The parameters of a distribution may refer to the output of functions via
+`function_name.output_name`. As a shorthand notation, when referring to the
+result of a single-output function, `function_name` may be used as an
+equivalent to `function_name.value`. So `"mean":"sum_1"` is equivalent to
+`"mean":"sum_1.value"`.
 
 Distributions can be treated either as `extended` or as non-`extended` [@barlow-extended-ml]. Some distributions are always extended, others can never be extended, yet others can be used in extended and non-extended scenarios and have a switch selecting between the two. An non-extended distribution is always normalized to the unity. An extended distribution, on the other hand, can yield values larger than unity, where the yield is interpreted as the number of predicted events. That is to say, the distribution is augmented by a factor that is a Poisson constraint term for the total number of events. 
 In the following, all distributions supported in HS${}^3$ v0.2.9 are listed in detail. Future versions will expand upon this list. 
--- a/docs/chapters/2.1_distributions.md
+++ b/docs/chapters/2.1_distributions.md
@@ -35,5 +35,13 @@ result of a single-output function, `function_name` may be used as an
 equivalent to `function_name.value`. So `"mean":"sum_1"` is equivalent to
 `"mean":"sum_1.value"`.
 
+The parameters of a distribution may also refer to variates of other
+distributions via `distribution_name.variate_name`. For example, if
+distribution `exp1` has a variate `x`, then the value of that variate can be
+referred to as `dist1.x`, e.g. `"sigma":"exp1.x"`, resulting in a hierarchical
+distribution (a folding of distributions). In the current version of this
+standard, this is an optional feature and may not be supported by all
+implementations. It may become mandatory in future versions.
+
 Distributions can be treated either as `extended` or as non-`extended` [@barlow-extended-ml]. Some distributions are always extended, others can never be extended, yet others can be used in extended and non-extended scenarios and have a switch selecting between the two. An non-extended distribution is always normalized to the unity. An extended distribution, on the other hand, can yield values larger than unity, where the yield is interpreted as the number of predicted events. That is to say, the distribution is augmented by a factor that is a Poisson constraint term for the total number of events. 
 In the following, all distributions supported in HS${}^3$ v0.2.9 are listed in detail. Future versions will expand upon this list. 
--- a/docs/chapters/2.2_functions.md
+++ b/docs/chapters/2.2_functions.md
@@ -35,6 +35,13 @@ equivalent to `function_name.value`. So
 `"summands" : [1.8, "sum_2", "param_xy"]` is equivalent to 
 `"summands" : [1.8, "sum_2.value", "param_xy"]`.
 
+The inputs of a function may also refer to variates of distributions via
+`distribution_name.variate_name`. So if  example, if a distribution `exp1` has
+a variate `x`, then the value of that variate can be referred to as `dist1.x`,
+e.g. `"summands" : [1.8, "exp1.x", "param_xy"]`. In the current version of
+this standard, this is an optional feature and may not be supported by all
+implementations. It may become mandatory in future versions.
+
 
 ### Product 
 A product of values or functions $a_i$. 
--- a/docs/chapters/2.2_functions.md
+++ b/docs/chapters/2.2_functions.md
@@ -36,7 +36,7 @@ equivalent to `function_name.value`. So
 `"summands" : [1.8, "sum_2.value", "param_xy"]`.
 
 The inputs of a function may also refer to variates of distributions via
-`distribution_name.variate_name`. So if  example, if a distribution `exp1` has
+`distribution_name.variate_name`. So for example, if a distribution `exp1` has
 a variate `x`, then the value of that variate can be referred to as `dist1.x`,
 e.g. `"summands" : [1.8, "exp1.x", "param_xy"]`. In the current version of
 this standard, this is an optional feature and may not be supported by all
--- a/docs/chapters/2.2_functions.md
+++ b/docs/chapters/2.2_functions.md
@@ -9,19 +9,32 @@ The top-level component `functions` describes an array of mathematical functions
 -   `name`: custom unique string 
 -   `type`: string that determines the kind of function, e.g. `sum` 
 -   `...`: each function has individual parameter keys for the various     individual parameters. For example, functions of type `sum` have the     parameter key `summands`. In general, these keys can describe     strings as references to other objects or numbers. Depending on the     parameter and the type of function, they appear either in single     item or array format. 
+-   `output`: label(s) the output(s) of the function. For single-output
+-   functions, this field is optional and set to `"value"` by default. A
+-   single output name (like `"output" : "value"`) and an array with a single output
+-   name (like `"output" : ["value"]`) are equivalent. For a function with
+-   multiple output values, `output` must be set to an array of names
+-   (like `"output" : ["a", "b", "c"]`).
 
 ```json title="Example: Functions"
 "functions": [ 
 	{ 
 		"name" : "sum1", 
 		"type" : "sum", 
-		"summands" : [1.8, 4, "param_xy"] 
+		"summands" : [1.8, "sum_2", "param_xy"],
+		"output" : "value"
 	}, 
 	... 
 ]
 ``` 
 
-In the following the implemented functions are described in detail. 
+The inputs of a function may refer to the output of other functions via
+`function_name.output_name`. As a shorthand notation, when referring to the
+result of a single-output function, `function_name` may be used as an
+equivalent to `function_name.value`. So
+`"summands" : [1.8, "sum_2", "param_xy"]` is equivalent to 
+`"summands" : [1.8, "sum_2.value", "param_xy"]`.
+
 
 ### Product 
 A product of values or functions $a_i$. 


docs/chapters/2.1_distributions.md
```diff
@@ -15,19 +15,33 @@ Each distribution [must]{.smallcaps} have the components `type`, denoting the ki
 	{ 
 		"name":"gauss1", 
 		"type":"gaussian_dist", 
-		"mean":1.0, 
+		"mean":"sum_1",
```
kratsg, line 18: this shouldn't be changed as it demonstrates outputting a singular value

docs/chapters/2.1_distributions.md
```diff
@@ -15,19 +15,33 @@ Each distribution [must]{.smallcaps} have the components `type`, denoting the ki
...
+`"mean":"sum_1.value"`.
+
+The parameters of a distribution may also refer to variates of other
+distributions via `distribution_name.variate_name`. For example, if
+distribution `exp1` has a variate `x`, then the value of that variate can be
+referred to as `dist1.x`, e.g. `"sigma":"exp1.x"`, resulting in a hierarchical
+distribution (a folding of distributions). In the current version of this
+standard, this is an optional feature and may not be supported by all
+implementations. It may become mandatory in future versions.
+
```
kratsg, line 45: I'm not sure of the need for this. The standard already allows for the cross-referencing as all variable names are meant to be globally shareable. This starts opening up a HUGE can of worms in complexity because now you have to start requiring a scope for every parameter specified.

docs/chapters/2.2_functions.md
```diff
@@ -9,19 +9,39 @@ The top-level component `functions` describes an array of mathematical functions
 -   `name`: custom unique string 
 -   `type`: string that determines the kind of function, e.g. `sum` 
 -   `...`: each function has individual parameter keys for the various     individual parameters. For example, functions of type `sum` have the     parameter key `summands`. In general, these keys can describe     strings as references to other objects or numbers. Depending on the     parameter and the type of function, they appear either in single     item or array format. 
+-   `output`: label(s) the output(s) of the function. For single-output
+-   functions, this field is optional and set to `"value"` by default. A
+-   single output name (like `"output" : "value"`) and an array with a single output
+-   name (like `"output" : ["value"]`) are equivalent. For a function with
+-   multiple output values, `output` must be set to an array of names
+-   (like `"output" : ["a", "b", "c"]`).
```
kratsg, line 17: Why is the name of the function not just used as the output? Why have a separate variable to split it off?

docs/chapters/2.2_functions.md
```diff
@@ -9,19 +9,39 @@ The top-level component `functions` describes an array of mathematical functions
...
-In the following the implemented functions are described in detail. 
+The inputs of a function may refer to the output of other functions via
+`function_name.output_name`. As a shorthand notation, when referring to the
+result of a single-output function, `function_name` may be used as an
+equivalent to `function_name.value`. So
+`"summands" : [1.8, "sum_2", "param_xy"]` is equivalent to 
+`"summands" : [1.8, "sum_2.value", "param_xy"]`.
+
+The inputs of a function may also refer to variates of distributions via
+`distribution_name.variate_name`. So if  example, if a distribution `exp1` has
```
AntonReinhard: typo
```suggestion
`distribution_name.variate_name`. So for example, if a distribution `exp1` has
```

docs/chapters/2.1_distributions.md
oschulz: We can put it somewhere else - but the other distribution example `exp1` already shows how to set a distribution parameter to a literal value. Since `"exp1"` is semantically a prior distribution (it's variate being called `param_x`), I think the use of a literal value fits better there, and the use of a computed parameter better into `gauss1`.

docs/chapters/2.1_distributions.md
oschulz: We absolutely need the ability to denote hierarchical/folded distributions. most importantly for priors in Bayesian use cases. Some of the demo use cases we have planned for HS3 absolutely require it, and other people involved also stated strongly that they need this.

docs/chapters/2.2_functions.md
oschulz: We need to support functions with multiple outputs for upcoming use cases, and we need to be able to route their outputs to different inputs. This, to me, seems the cleanest way to do this while keeping backward compatibility. It also resolves a perceived difference in the role of a function name vs. a distribution name, that several people commonted on negatively at the meeting this week (caused quite a bit of confusion): for distributions, the name has nothing to do with the variate names (which are equivalent to "output" names in PPL semantics - what sampling produces - and they are what the data names are matched to). Going forward like this, it's unified, clear, and - I believe - very user-friendly.

kratsg: Yes, but you can just use `function_name[0]` etc if you wanted. Why do we need two names referring to the same function?

docs/chapters/2.1_distributions.md
kratsg: Can you provide an example that's not currently supported by the spec as it is for the hierarchical/folded distributions? I fail to see why the scope is explicitly needed.

docs/chapters/2.1_distributions.md
kratsg: You changed this though. `gauss1` had `param_x` which treated `gauss1` as the prior. We shouldn't change out existing examples. If you want to show or demonstrate a different example for a different use case, it should be added - but we shouldn't change existing examples.

oschulz: Yes, since I felt `exp1` was more suitable in a prior role, esp. since `gauss1` is the first example, and the frequentist users will only used distributions over data, not parameters.

But if people feel strongly that this should not be changed, I'll be happy with another variant of the example distributions.

@cburgard ?

docs/chapters/2.1_distributions.md
oschulz: This prior for example (in paper/PPL syntax):

```
average_baseline_level ~ Normal(0.0, 3.0)
baseline_variation ~ Exponential(1.5)
baseline1 ~ Normal(averabe_baseline_level, baseline_variation)
baseline2 ~ Normal(averabe_baseline_level, baseline_variation)
baseline3 ~ Normal(averabe_baseline_level, baseline_variation)

priordist = productdist(average_baseline_level, baseline_variation, baseline1, baseline2, baseline3)
```

Constructions like this are very common in Baysesian use cases where prior parameter correlations go beyond simple Guassian covariances.

docs/chapters/2.2_functions.md
oschulz: I think we absolutely should *not* refer to output field names via a numerical index. For once, it's completely inconsistent with distributions, which are routed/matched to data by variable name. And in general, HS3 refers to things by name. Also, we want to be able to use tables (e.g. via Arrow) as universal input/output representations, and that means multi-output functions returning named tuples (semanticalls) so that multiple values become tables with column names. Lastly, we do need to support vector-values in the not too far future, and that means we need indices for that.

In the whole of HS3, things have names, to multi-output functions should return named tuples, not plain tuples. Having "value" as the default entry allows us to bring single-output functions into this in a clean, non-breaking fashion.

docs/chapters/2.1_distributions.md
kratsg: Yes, but in the above example, this is done already via

```json
  "distributions": [
    {
      "name": "average_baseline_level",
      "type": "normal",
      "mean": 0.0,
      "sigma": 3.0
    },
    {
      "name": "baseline_variation",
      "type": "exponential",
      "rate": 1.5
    },

    {
      "name": "baseline1",
      "type": "normal",
      "mean": "average_baseline_level",
      "sigma": "baseline_variation"
    },
    {
      "name": "baseline2",
      "type": "normal",
      "mean": "average_baseline_level",
      "sigma": "baseline_variation"
    },
    {
      "name": "baseline3",
      "type": "normal",
      "mean": "average_baseline_level",
      "sigma": "baseline_variation"
    }
  ],

  "prior": {
    "type": "product_distribution",
    "distributions": [
      "average_baseline_level",
      "baseline_variation",
      "baseline1",
      "baseline2",
      "baseline3"
    ]
  }
```

what's the problem missing? This is covered in the spec. So I'm not sure I understand.

szabo137:
I would like to raise the following concern: 

This notation could break with the JSON file format; In this case, no JSON reader/visualizer/parser can handle that automatically. I know (also from online discussions) that the parsing overhead is small, but I guess no established tool is willing to do that. This could cause overhead in debugging HS3 files, because some connections (from the concrete output of a function to inputs elsewhere) are not encoded in JSON, but please correct me if I'm wrong. 

Therfore, I would like to understand, where the dot-notation is used to consume the (maybe partial) output of a function? Just in the expression of other functions (which are interpreted on the engine side anyway), or also in "defining structures" like names and variates?

docs/chapters/2.1_distributions.md
oschulz: But that's not really consistent, because we always and explicitly give the variates of distributions, even univariate distributions, a name - using the distributions name instead when referring to it, to me, only makes sense if we say that `"sigma": "baseline_variation"` is a shorthand for `"sigma": "baseline_variation.x"`. 

Otherwise, how do we route variates of multivariate distributions? For example:

```
[fraction1, fraction2, fraction2] ~ Dirichlet([1.0, 1.0, 1.0])
total_momentum ~ Exponential(4.2)
momentum_fraction1 = fraction1 * total_momentum
momentum_fraction2 = fraction2 * total_momentum
momentum_fraction3 = fraction3 * total_momentum
```

Also, we saw at the meeting how people got very confused by the fact that for distributions, the variate names seemed to matter and for functions, the function name. With the proposal here that is both clarified and at the same time we gain the additional expressiveness we need.

kratsg: > Also, we saw at the meeting how people got very confused by the fact that for distributions, the variate names seemed to matter and for functions, the function name.

Was this on Day 2? I wasn't around for it, so it would be good to have the additional context here in GitHub also for other people's benefits who are not part of DEMOS or were not able to attend.

-----

Can you point at an existing implementation of `Dirichlet` where the pdf (or logpdf) evalutes to multiple outputs? [pymc](https://www.pymc.io/projects/docs/en/stable/api/distributions/generated/pymc.Dirichlet.html) and [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.dirichlet_multinomial.html) both treat this as a $f : \mathbb{R}^N \to \mathbb{R}$. So I'm curious to see an implementation that is $f : \mathbb{R}^N \to \mathbb{R}^M$.

I think the confusion is you're trying to represent sampling in HS3, but HS3 as I understand is about composition of distributions and functions... or rather, with this pseudocode (thinking in terms of pymc)

```python
import pymc as pm

with pm.Model() as model:

    fractions = pm.Dirichlet(
        "fractions",
        a=[1.0, 1.0, 1.0],
    )

    T = pm.Exponential(
        "total_momentum",
        lam=4.2,
    )

    momentum = pm.Deterministic(
        "momentum",
        fractions * T,
    )
```

which is what a user does (and what I think you're proposing HS3 is meant to serialize or capture). But `pymc` internally will take the above and give

```python
logp(theta):
    logp = 0

    logp += dirichlet_logpdf(fractions, a)
    logp += exponential_logpdf(T, lam)

    momentum = fractions * T

    return logp
```

and this is what is getting serialized.

oschulz:
> I would like to raise the following concern:
> 
> This notation could break with the JSON file format; In this case, no JSON reader/visualizer/parser can handle that automatically.

We already allow for mathematical expressions in `generic_function`, which we'll need to parse during HS3 validation, and that can be more complex than the output/variate reference syntax proposed here.

> I know (also from online discussions) that the parsing overhead is small, but I guess no established tool is willing to do that.

A plain regular expression like `^[a-zA-Z_][a-zA-Z0-9_]*(?:\.[a-zA-Z_][a-zA-Z0-9_]*)?(?:\[\d+(?:\s*,\s*\d+)*\])?$` should do it quite nicely - this one would cover `name`, `name.value` and even a future `name.value[index1, index2, ...]`. So I don't think that's an issue.

> This could cause overhead in debugging HS3 files, because some connections (from the concrete output of a function to inputs elsewhere) are not encoded in JSON, but please correct me if I'm wrong.

I would argue that it would provide a huge benefit when reading/debugging JSON files manually because it'll make the flow of values and variates much easier to see compared to explicit lens/traversal nodes that could bloat the files quite a bit (and would make writing HS3 snippets with multi-output functions a pain).

We could, of course do something like `"input": "some_function"` for the single-output case and `"input": {"source": "some_function", "field": "some_output"}` for the multi-output case - and later, when we'll add vector/array values, `"input": {"source": "some_function", "field": "some_output", "indices": [3, 4]}`. But to me that seems harder to read, much harder to type, than `some_function.some_output[3, 4]`, just to avoid a single regular expression in the engines.

> Therfore, I would like to understand, where the dot-notation is used to consume the (maybe partial) output of a function? Just in the expression of other functions (which are interpreted on the engine side anyway), or also in "defining structures" like names and variates?

The dot-notation would be used everywhere (but we'll allow non-dot as a shorthand for functions with a single output named "value") where an input of a computational DAG node is connected to an output in the PPL view. In a PPL, when sampling, the inputs of a (multivariate) distribution with free parameters (so technically a Markov kernel) are those free parameters. An the PPL outputs of a distribution are the fields of a random value drawn from it. So a DAG that supports both function and distribution (Markov kernel) nodes, with potentially multiple inputs and outputs, an edge is naturally, `src.some_output-> dest.some_input`, which in HS3 JSON becomes `{"name": "dest", "type": "...", ..., "some_input" : "src.some_output"}`, not matter if `dest` is a function or a distribution.

HS3, at it's heart, has always been (and needs to be) a kind of "document-formatted PPL", and we should treat it as such.

In any case it's important to have names for everything so we can use tables as the universal I/O primitive, for remote models, disk I/O, etc., which gives us a single powerful abstraction. Tables need column names, so even for a function with a single output, that output needs a name. By using `value` as a default, people won't have to deal with the dot syntax where the semantics are clear, but on the other hand we also clearly define what an output table will look like.

docs/chapters/2.1_distributions.md
oschulz: > Was this on Day 2?

Sorry, yes - we had long in-depth discussion on this, because several people found the current way of referring to things confusing (even I confused myself at some point and I helped write the spec). And several people clearly stated that both multi-output functions and hierarchical models are necessary to them.

> Can you point at an existing implementation of Dirichlet where the pdf (or logpdf) evalutes to multiple outputs?

No, of course not. The "output" (in the PPL-DAG sense) of a distribution it not it's (log-)PDF, it is it's variate(s). And it is in the PPL-sense that people need to connect that "PPL-output" of a distribution to the inputs of function and free parameters of other distributions to build hierarchical models. The PDF plays a role for the likelihood objects, not for connecting functions and distributions. And so we would refer to `gauss1.x`, in another function/distribution node, not to it's PDF.

> I think the confusion is you're trying to represent sampling in HS3

Sampling absolute is an important way people will use HS3, for example to generate synthetic/toy/test datasets. And they will also want to run it in "PPL forward mode", so to speak, just with "mean" (analytically or stochastically) instead of "rand", to generate Asimov datasets. Being able to sample is also vital to enable simulation-based inference.

In inference mode, it is the likelihood objects that users will access primarily, not distribution objects.

> but HS3 as I understand is about composition of distributions and functions... or rather, with this pseudocode (thinking in terms of pymc)

HS is not just an "inference machine". It describes stochastic forward models - essentially Markov kernels that map parameters to probability distributions/measures over data and data, and can contain (but doesn#t have to) pre-defined likelihoods that combines models with data. The forward models, seen as a PPL DAG, are often comprised of a graph of deterministic nodes (functions) with a final layer of probabilistic nodes (distributions with free parameters).

The total graph, when it models data, will often have free input - typically the fit parameters. A graph that describes a prior, on the other hand, will have no "open" inputs. In the prior case it is vital that the graph may also contain probabilistic nodes in intermediate layers, so that we can have hierarchically correlated priors. In the case of models refer to data, this will be less common, since it will often make this likelihood intractable. The will definitely be users that need such models though and who will then use methods like simulation-based inference to deal with that.

cburgard: Hey everyone, 
I'm not trying to stop the discussion (please continue) but I feel like there's a potential misunderstanding that needs to be cleared up before you proceed.

In @kratsg s mental model (and mine), the output of a distribution is the probability value at the given data point (RooFit-speak `getVal`).
In @oschulz s mental model, the output of a distribution is a random sample of the variate(s) (RooFit-speak `generate( ... )`, where `...` is one or more of the variates).

I think this proposal is mostly about a way to refer to the variates of a distribution in a generative context, that is "x as generated by mygauss". This is needed, as the same `x` can be the variate (input or output, depending on your point of view) of several distributions. 

I think the specifics of the notation (whether a dot is appropriate or not) is fore sure still up for debate (this is why we are here), but I at least am not aware of a way to uniquely refer to the variate of a distribution, as HS3 does allow a different distribution to link to the same variate at the same time.


```
"distributions": [

    {
      "name": "mygauss",
      "type": "normal",
      "mean": 0,
      "sigma": 1,
      "x": "x"
    },

    {
      "name": "mypoisson",
      "type": "poisson",
      "lambda": "lambda",
      "x": "x"
    }
]
```

cburgard: Reflecting on this further, though, I think it would be most useful to discuss this point with a concrete use case in hand. I do see the point that HS3 is meant to describe the model (distributions and functions), not neccesarily an algorithm that is performed by the model (sampling, minimization) - but potentially there are cases where these two cannot be so cleanly separated...

oschulz: > I do see the point that HS3 is meant to describe the model (distributions and functions), not neccesarily an algorithm that is performed by the model (sampling, minimization) - but potentially there are cases where these two cannot be so cleanly separated

Well, one of our main points is that we do want to separate model and inference, right? :-)

oschulz: > In @kratsg s mental model (and mine), the output of a distribution is the probability value at the given data point (RooFit-speak getVal).
> In @oschulz s mental model, the output of a distribution is a random sample of the variate(s) (RooFit-speak generate( ... ), where ... is one or more of the variates).

Indeed - a distribution itself doesn't really have an output, of course. A probability measure does, but it's input is a set, not a point.

So when I said "output" I always meant output in a PPL-DAG sense.

But no matter what's it's called:

* We do need functions with multiple named outputs, and inputs of functions and (for hierarical models/priors) parameters of distributions need to refer to them, and refer to them selectively.

* We already have distributions with multiple named variates, and for hierarical models/priors, the inputs of functions and parameters of distributions need to refer to them, and refer to them selectively.

So I think allowing `source_function.output_name` and `source_distribution.variate_name` in the inputs of functions and parameters of distributions is

* Intuitive for users (after all, the variate named referred to are set explicitly in the distributions already, and we will set them explicitly in multi-output functions, and `"value"` is a sensible default for single-output functions).

* Easy to read, both by humans and machines.

* Consistent.

In the case of composite distributions, there's a duality in regard to whether we refer to the (log)-PDF of another distribution or it's variates:

* The `dist` field of `rate_extended_dist` (our inhomogenous Poisson point process) could be read as: "Use the PDF of that distributions as the normalized Poisson expectation density here". But one can also read it as a model `n ~ Poisson(rate), x ~ dist^n`, which refers to the variate.

* For product distributions, it's PDF is the direct procuct of the PDFs of the components, but it's variate is the cartesian product of their variate.

So I think for composite distributions, they should refer to other distributions by name only, as before.

szabo137:
> We already allow for mathematical expressions in `generic_function`, which we'll need to parse during HS3 validation, and that can be more complex than the output/variate reference syntax proposed here.

I know, and I am fine with that, and also if specific output values are referred to using the dot-notation *in an expression*. However, my concern is that we introduce some complications if we use the dot-notation anywhere else in the HS3 file. 

> A plain regular expression like `^[a-zA-Z_][a-zA-Z0-9_]*(?:\.[a-zA-Z_][a-zA-Z0-9_]*)?(?:\[\d+(?:\s*,\s*\d+)*\])?$` should do it quite nicely - this one would cover `name`, `name.value` and even a future `name.value[index1, index2, ...]`. So I don't think that's an issue.

This is obviously no issue in any code we own. However, in every code we don't own (reader/parser/visualizer of JSON files), these fields are just string-blobs without any "JSON-compliant" connection to other parts of the HS3 file. The connection-logic between input and output, which is given in the current HS3 standard by the JSON structure, will then be hidden behind string blobs, which are just values of the fields in the file.


> I would argue that it would provide a huge benefit when reading/debugging JSON files manually because it'll make the flow of values and variates much easier to see compared to explicit lens/traversal nodes that could bloat the files quite a bit (and would make writing HS3 snippets with multi-output functions a pain).

With this, you have a point when it comes to human-readability. However, the question is then, what does human-readability actually mean in this context? If I open the HS3-file in a text editor, you are right, the dot-notation is short and easy to follow. But if you open the file in a JSON-viewer (what people tend to do), everything is just chunked together in one field "input". Then resolving that `{"name": "func", "out": ["vec_value", "scalar_value"]}` becomes `"func.vec_value[2,3]"` somewhere else could be challenging in a megabyte JSON file. In this regard, the human-readability argument is questionable. Furthermore, IMHO, I think "educational presentability" is not a valid argument when it comes to the design of a serialization standard. 

And just to make clear, I didn't talk about any "lens/traversal nodes"; this was a quick and stupid idea in our offline discussion. Here, I would like to talk about the *syntax* for adding multiple outputs and solving the mismatching meaning of the field "name". 

> We could, of course do something like `"input": "some_function"` for the single-output case and `"input": {"source": "some_function", "field": "some_output"}` for the multi-output case - and later, when we'll add vector/array values, `"input": {"source": "some_function", "field": "some_output", "indices": [3, 4]}`. But to me that seems harder to read, much harder to type, than `some_function.some_output[3, 4]`, just to avoid a single regular expression in the engines.

IMHO, I think this is the best solution. It is somewhat human-readable, it is easy to extend, and it preserves the logic of the JSON-format. Maybe I'll post an example in another comment. 


> The dot-notation would be used everywhere (but we'll allow non-dot as a shorthand for functions with a single output named "value") where an input of a computational DAG node is connected to an output in the PPL view. In a PPL, when sampling, the inputs of a (multivariate) distribution with free parameters (so technically a Markov kernel) are those free parameters. An the PPL outputs of a distribution are the fields of a random value drawn from it. So a DAG that supports both function and distribution (Markov kernel) nodes, with potentially multiple inputs and outputs, an edge is naturally, `src.some_output-> dest.some_input`, which in HS3 JSON becomes `{"name": "dest", "type": "...", ..., "some_input" : "src.some_output"}`, not matter if `dest` is a function or a distribution.

Again, I do not have any worries with the added logic, and I think unifying this input and output format of distributions and functions in an HS3 file would be very beneficial for various reasons. I just have issues with the syntax you want to introduce to add this new logic. 

> HS3, at it's heart, has always been (and needs to be) a kind of "document-formatted PPL", and we should treat it as such.

Fully agree with that. And the description of the standard should reflect that in whatever human-readable language, like your dot-notation. However, HS3 is currently somewhat bound to the JSON file format, and breaking with that by introducing a syntax that is not JSON-compatible leads this decision to absurdity. 

> In any case it's important to have names for everything so we can use tables as the universal I/O primitive, for remote models, disk I/O, etc., which gives us a single powerful abstraction. Tables need column names, so even for a function with a single output, that output needs a name. By using `value` as a default, people won't have to deal with the dot syntax where the semantics are clear, but on the other hand we also clearly define what an output table will look like.

Again, I fully agree with that, I just think it is a better idea to reflect this in a JSON-compatible way instead of introducing yet another syntax to a file which already has at least two: JSON and the *expressions*.

szabo137:
As I promised, here is an example to compare the proposed dot-notation (as I understand it) and a JSON-compatible version: 

### version 1 (dot-notation)

```json
{
      "name": "func_with_dot",
      "input": [
        "some_function.some_output[3,4]",
        "some_function2.some_output2",
        "some_function2.some_other_output2"
      ],
      "output": ["val1", "val2", "val3" ]
    }
```
Rendered like 

Screenshot 2026-03-04 at 21 28 55

Here one sees that the string-blobs like `"some_function.some_output[3,4]"` is the short-hand dot-syntax for `{"name": "some_function", "output":["some_output", ...]}` somewhere else. 


### version 2 (JSON-compatible)

```json
{
      "name": "func_without_dot",
      "input": [
        {
          "source": "some_function",
          "fields": "some_output",
          "indices": [3, 4]
        },
        {
          "source": "some_function2",
          "fields": ["some_output2","some_other_output2"]
        }
      ],
      "output": ["val1", "val2","val3"]
    }
```
Rendered like 

Screenshot 2026-03-04 at 21 28 16

Here, the JSON-structure of the output of one function, e.g., 

```json
{"name": "some_function", "output":["some_output", ...]}
```

is structurally similar to the input of the other function:

```json
"input": {"source": "some_function", "fields": "some_output", "indices": [3, 4]}
```
Maybe one should actually use other words for *source* and *fields*, maybe *name* and *values*: 

```json
"input": {"name": "some_function", "values": "some_output", "indices": [3, 4]}
```

kratsg:
The dot notation will break generic parsing for pretty much every library in existence that I'm aware of. I would rather not create a brand new grammar. I would say that if tools like Mathematica and SymPy are unable to parse the expression, then we have bigger problems. Even getting this implemented in ROOT will be an uphill battle, because since you allow numbers as part of the names (e.g. `varx2.4` could be seen as either `varx*2.4` or `varx2.4` and this is not great, unless you only allow `.` in between alphabetical characters when it's part of a variable name, and this gets very strained.

Instead, what is going to need to be done is how ROOT has generally approached this, and that would be through aliases where you have a separate look-up table to map through outputs of functions and inputs of others.

oschulz:
> ### version 1 (dot-notation)
> ### version 2 (JSON-compatible)

Personally, I find version 1 more readable, even in a JSON viewer.

I'm not saying we can't do version 2 - though it should be `"field"`, singular, multi-output to single input could get messy, I think - but I really don't quite see the benefit. Which concrete tools, not written as part of HS3, could so what better with the JSON-structured syntax?

oschulz:
> The dot notation will break generic parsing for pretty much every library in existence that I'm aware of.

I really don't quite see how? The JSON will be parsed, and how we interpret a name-field, with our without inner structure, is up to us in the end.

> I would rather not create a brand new grammar. I would say that if tools like Mathematica and SymPy are unable to parse the expression, then we have bigger problems.

This is about connecting nodes in the graph, and HS3 structure. Why would we give a reference string like that directly to Mathematica or SymPy? They can't understand the higher level structure at all anyway.

Or do you mean in "free-form" mathematical expressions? I would suggest to re-write `a.b` to something like `a__b` there internally, and insure `a.b` get's routed to it (just one possibility).

If we do not allow for `a.b` references at all in HS3 "free-form" math, then we will have to have explicit lens nodes in the graph. We can do that, of course - I just think it will make HS3 files a lot less readable.

 > Even getting this implemented in ROOT will be an uphill battle, because since you allow numbers as part of the names (e.g. `varx2.4` could be seen as either `varx*2.4` or `varx2.4`

I don't think so - we clearly specify that dots are not allowed to be part of HS3 object names, and that names must not start with a number (a common convenion for variable/function/... names in programming languages, after all). So `varx2.4` would not be legal HS3 in any case. A dot would clearly separate an object name from an output/variate name, just like it separates an object names from a field name in many languages. I don't think there's any possibility of ambiguity.

kratsg:
> I don't think so - we clearly specify that dots are not allowed to be part of HS3 object names, and that names must not start with a number (a common convenion for variable/function/... names in programming languages, after all). So `varx2.4` would not be legal HS3 in any case. A dot would clearly separate an object name from an output/variate name, just like it separates an object names from a field name in many languages. I don't think there's any possibility of ambiguity.

This still requires a brand new grammar in pretty much every single expression parser. All of this would be better suited in a [discussion](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/discussions) rather than as a PR because there's too many unknowns to deal with and lots of varying issues that needed to be dealt with from a technical perspective before trying to implement it in the specification. For example, [ROOT has the issue of masking variable names](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/discussions/45) and that's just the primary implementation we've been dealing with when it comes to HS3.

In the above example, I can already tell you that by making it python-specific while also failing to be parseable in general python grammar, you make it impossible to use C++ or Julia's grammar (AST) for it, that is

```
        "some_function.some_output[3,4]",
```

is not valid python. It's not clear if you mean this to represent `[3:4]` or `[3][4]`, or `[(3,4)]` etc... There's an Array API but that's not universal. Let's suppose you want the expression to be something like

```
"expression": "some_function.some_output[3,4] * x + 159"
```

following the support of this input/output, the problem then is you can't actually write this in the JSON.. what you would need to be writing in the JSON instead is the intermediate representation (IR) as

```json
{
  "type": "add",
  "lhs": {
    "type": "mul",
    "lhs": {
      "type": "ref",
      "function": "some_function",
      "output": "some_output",
      "index": [3,4]
    },
    "rhs": { "type": "ref", "name": "x" }
  },
  "rhs": { "type": "const", "value": 159 }
}
```

so that every language goes from IR->execution. Otherwise, you're going to fight this forever across different languages. So you would then instead adopt this via ONNX syntax which is more like what @szabo137 mentioned in the second example and is actually far far easier to implement than the first example...

oschulz:
There are clearly major concerns from some of the main implementor of the standard and I think I've made all the arguments I can, so let's move on.

Closing this in favor of #86.

PR #86, opened 2026-03-05
Add support for named-tuple functions and array values

oschulz:
An alternative approach to #84.

* Clarify the role of named tuples.

* Add support for passing around array values.

* Add lens functions `get_property`, `get_properties`, `get_element`, `get_elements`.

* Add named tuple constructor function `named_tuple`.

With this approach, the output of a function and the variate of a distribution is semantically always a single output, so all nodes in the PPL DAG are single-output (but that output may fan out to multiple inputs, of course).

This approach is also non-breaking but provides clear unified semantics on the role of function names vs. distributions names that have apparently been confusing to people so far.

It should also hopefully address

> @kratsg : what is going to need to be done is how ROOT has generally approached this, and that would be through aliases where you have a separate look-up table to map through outputs of functions and inputs of others.

in a semantically clean and PPL-DAG compatible fashion.

This PR already adds array support - we may as well do this now, so we can see the full picture.


kratsg:
Do you mind opening up a discussion instead? https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/discussions -- this allows multiple threads to spawn until the same discussion on various key points which is easier to keep track of.

oschulz:
> Do you mind opening up a discussion instead?

Well, this is a concrete proposal, it might be easier to discuss here, where the source is available?

kratsg:
> Well, this is a concrete proposal, it might be easier to discuss here, where the source is available?

You can link the source code from there too, but you can't make threads in PRs.

docs/chapters/2.2_functions.md
```diff
@@ -15,13 +15,175 @@ The top-level component `functions` describes an array of mathematical functions
...
 	... 
 ]
 ``` 
 
-In the following the implemented functions are described in detail. 
+Any input of a function may refer to the output of other function via
+it name, as in `"some_arg" : "some_function"`,
+`"summands" : [..., "some_function", ...]` and similar.
+
+The inputs of a function may also refer to the variate of a distributions via
```
szabo137:
```suggestion
The inputs of a function may also refer to the variate of a distribution via
```

docs/chapters/2.2_functions.md
```diff
@@ -15,13 +15,175 @@ The top-level component `functions` describes an array of mathematical functions
...
+		"type" : "get_element",
+		"from" : "mv_normal2",
+		"select": 3
+	}, 
+	... 
+]
+```
+
+### Array multi-element accessor
+Extracts a multiple elements from an array and names them.
```
szabo137:
```suggestion
Extracts multiple elements from an array and names them.
```

docs/chapters/1_introduction.md
```diff
@@ -13,7 +13,21 @@ This document sets out to document the syntax and features of the Statistics Ser
...
+
+### Values
+
+HS${}^3$ currently supports these types of computational values:
+
+- real numbers
+
+- n-dimensional arrays of real numbers
+
+- named tuples of real numbers and n-dimensional arrays of real numbers: a named tuple `(some_property = ..., some_other_property = ..., ...)` is an unnamed struct with a fixed field order. It may realized as a named tuple (if supported by the programming language), a dictionary with fixed keys, a single row of a table with respective column names and so on in HS${}^3$ implementations.
```
szabo137:
```suggestion
- named tuples of real numbers and n-dimensional arrays of real numbers: a named tuple `(some_property = ..., some_other_property = ..., ...)` is an unnamed struct with a fixed field order. It may be realized as a named tuple (if supported by the programming language), a dictionary with fixed keys, a single row of a table with respective column names and so on in HS${}^3$ implementations.
```

docs/chapters/1_introduction.md
```diff
@@ -13,7 +13,21 @@ This document sets out to document the syntax and features of the Statistics Ser
...
+
+Outputs of functions may be numbers, arrays or named tuples, variates of distributions are always named tuples.
```
szabo137:
```suggestion
Outputs of functions may be numbers, arrays, or named tuples; variates of distributions are always named tuples.
```

docs/chapters/1_introduction.md
```diff
@@ -13,7 +13,21 @@ This document sets out to document the syntax and features of the Statistics Ser
...
+
+Complex numbers are currently not supported, but may be added in a future version of this standard.
```
szabo137:
```suggestion
Complex numbers are not currently supported, but may be added in a future version of this standard.
```

docs/chapters/2.1_distributions.md
```diff
@@ -10,23 +10,87 @@ Each distribution [must]{.smallcaps} have the components `type`, denoting the ki
...
 
-```json title="Example: Distributions"
+A parameter of a distribution may refer to the output of a function via its
+name.
+
+A parameter of a distribution may also refer to (the variate of) another
+distributions via its name. Note though that
+distribution parameters will often be scalar or arrays while distribution
+variates are always named tuples in this standard. So in practice, one will
+often use a function like `get_property` in between.
```
szabo137, line 20:
```suggestion
A parameter of a distribution may also refer to (the variate of) another
distribution via its name. Note, though, that
distribution parameters will often be scalar or arrays while distribution
variates are always named tuples in this standard. So in practice, one will
often use a function like `get_property` in between.
```

docs/chapters/2.2_functions.md
```diff
@@ -15,13 +15,175 @@ The top-level component `functions` describes an array of mathematical functions
...
-		"summands" : [1.8, 4, "param_xy"] 
+		"summands" : [1.8, "prod1", "param_xy"] 
 	}, 
 	... 
 ]
 ``` 
 
-In the following the implemented functions are described in detail. 
+Any input of a function may refer to the output of other function via
+it name, as in `"some_arg" : "some_function"`,
```
szabo137:
```suggestion
Any input of a function may refer to the output of another function via
it name, as in `"some_arg" : "some_function"`,
```

docs/chapters/2.2_functions.md
```diff
@@ -15,13 +15,175 @@ The top-level component `functions` describes an array of mathematical functions
...
+`"summands" : [..., "some_function", ...]` and similar.
+
+The inputs of a function may also refer to the variate of a distributions via
+its name, like in `"some_arg" : "some_distribution"`. Note though that
+function arguments will often be scalar and that distribution variates are
+always named tuples in this standard. So in practice, one will often use
+function like `get_property` in between (see below for an example).
```
szabo137:
```suggestion
its name, like in `"some_arg" : "some_distribution"`. Note, though, that
function arguments will often be scalar and that distribution variates are
always named tuples in this standard. So in practice, one will often use
a function like `get_property` in between (see below for an example).
```

oschulz:
Well, I guess it's a matter of preference, I've seen complex issues discussed and resolved on large projects via the PR format over the years. In any case, I'm leaving this PR open for now so people to comment, and invite interested parties to suggest improvements or alternatives approaches to the HS3 extensions we need via separate PRs or discussions.

szabo137 (approved):
LGTM! Just some typos. 

I think `get_property` is the cleanest direct solution. However, I think in the future, we might come up with a more implicit solution, but for now, I would be happy if we proceed with this 🙆‍♂️

oschulz:
Thanks for the corrections @szabo137 ! It's weird, I ran my PR through Claude for fixes before pushing, and it found and corrected some spelling/grammar errors, but it somehow not these, even though they're clear.

oschulz:
> I think get_property is the cleanest direct solution. However, I think in the future, we might come up with a more implicit solution, but for now, I would be happy if we proceed with this.

I think we'll need it `get_property` and friends long-term anyway, because they're easier to emit when auto-generating HS3. And like you say, human-friendly syntactic sugar can still be added later if people ask for it.

kratsg (commented):
"Named tuple" in Python means something very specific: `collections.namedtuple` or `typing.NamedTuple`, which are immutable, ordered, and importantly are tuple subclasses with positional access. In Julia `NamedTuple` is a built-in parametric type that maps more closely to this, but for a cross-language spec this is confusing. "Struct" or "record" would both be better. The spec itself even says "an unnamed struct with a fixed field order" in the definition. I'd suggest proposing either "record" (language-neutral, used in Avro/JSON Schema/database contexts) or "struct" (familiar from C/C++/Rust/Go). The function type named_tuple should follow suit (e.g., record or struct).

The `from` field creates a type-ambiguous resolution problem. Throughout the new accessor functions, from can be "a function or distribution name or a literal value." But the spec doesn't define how an implementer should disambiguate these at parse time. In Python you'd need to resolve a name against both the `functions` and `distributions` namespaces and also check if it's a literal. The spec should clarify the resolution order, or better yet, state explicitly that names must be unique across both functions and distributions (which I think is already implied elsewhere, but it's not restated here and it matters a lot for these accessor functions).

The `get_element` indexing is 1-based (right???). The example says "extract the third element" and uses `"select": 3`. This is fine-ish for Julia/Fortran but gets tedious in other languages that don't do indexing the same way such as Python/C++. The spec must explicitly state whether indices are 0-based or 1-based. This is a guaranteed source of off-by-one bugs across implementations if left ambiguous. Given that the serialization format is JSON (which uses 0-based arrays), I'd argue for 0-based and flag this as a required clarification.

`get_elements`: The spec says select can be `[i1, i2, ...]` for 1D or `[[i1, j1, ...], [i2, j2, ...], ...]` for multi-dimensional. But how does an implementer distinguish between "select elements `1, 2, 3` from a 1D array" `([1, 2, 3])` and "select element at index `(1, 2, 3)` from a 3D array" `([1, 2, 3])`? These are the same JSON. You'd need to inspect the dimensionality of the source array at runtime, which means the type system is no longer statically resolvable from the JSON alone. I'd suggest requiring the nested-array form always for multi-dimensional (i.e., `[[1], [2], [3]]` for 1D multi-select vs `[[1, 2, 3]]` for a single 3D index) which is already done in other places in the spec.

`get_properties`: is `as` optional? If I just want to extract a subset of properties without renaming, do I still need to provide `as`? The spec doesn't say whether `as` is required or optional.

Distribution variates are "always named tuples" but what about scalar distributions? A `gaussian_dist` has variate `x`. Is the variate of a Gaussian now a named tuple with one field? That's `(x = value)` rather than just `value`. This means that every consumer of a distribution output must go through `get_property` even for trivially scalar distributions. That's a lot of boilerplate in the JSON and a lot of wrapper objects in Python. I'd want the spec to clarify whether scalar distributions can have a scalar variate as a special case, or if the one-element named tuple is truly mandatory.

Overall, the accessor/constructor pattern (`get_property`, `get_properties`, `get_element`, `get_elements`, `named_tuple`) is essentially building a mini functional language for data plumbing which is generally already provided by things like ONNX for you with a well-defined schema... maybe we should be adopting ONNX schemas as subsets of the HS3 spec?

docs/chapters/2.2_functions.md
```diff
@@ -15,13 +15,175 @@ The top-level component `functions` describes an array of mathematical functions
...
+		"mean" : [...],
+		"covariance" : [...]
+	},
+	...
+],
+"functions": [ 
+	{ 
+		"name" : "values_d_e", 
+		"type" : "get_properties",
+		"from" : "mv_normal1",
```
kratsg, line 87: `mvnormal1`

docs/chapters/2.2_functions.md
```diff
@@ -15,13 +15,175 @@ The top-level component `functions` describes an array of mathematical functions
...
+		"mean" : [...],
+		"covariance" : [...]
+	},
+	...
+],
+"functions": [ 
+	{ 
+		"name" : "mvnormal2_A1", 
+		"type" : "get_element",
+		"from" : "mv_normal2",
```
kratsg, line 122: `mvnormal2`

docs/chapters/2.2_functions.md
```diff
@@ -15,13 +15,175 @@ The top-level component `functions` describes an array of mathematical functions
...
+
+-   `name`: custom unique string 
+-   `type`: `get_elements` 
+-   `from`: a name that refers to a function or distribution or a literal value.
+-   `select`: A array of integer indices `[i1, i2, ...]` or (for
+    multi-dimensional arrays) an array of arrays of integer indices
+    `[[i1, j1, ...], [i2, j2, ...], ...]`.
+-   `as`: An array of new names for the extracted elements.
+  
+Example (extract the third element of the variate of a multivariate Normal
```
kratsg, line 140: copy/paste problem?

docs/chapters/2.1_distributions.md
```diff
@@ -10,23 +10,87 @@ Each distribution [must]{.smallcaps} have the components `type`, denoting the ki
...
+	},
+	...
+],
+"functions": [ 
+	{ 
+		"name" : "average_baseline_level", 
+		"type" : "get_property", 
+		"from" : "avg_bl_dist", 
+		"select" : "average_baseline_level" 
+	},
```
kratsg, line 85: The function `"average_baseline_level"` extracts property `"average_baseline_level"` from `"avg_bl_dist"`. But `avg_bl_dist` is a Gaussian with `"x": "average_baseline_level"`. So the variate of the distribution is a struct with field `average_baseline_level`, and the `get_property` extracts that field. This is really verbose for scalar distributions. Every scalar prior requires both a distribution and a companion `get_property` function just to unwrap it.

kratsg (changes requested):

docs/chapters/2.1_distributions.md
oschulz: I fully agree, this would be quite painful. We may want to allow for distributions that have plain values as variates instead of records.

oschulz:
> "Named tuple" in Python means something very specific

Yes I've also not been that happy with the term. I propose to change this to "record" - the proper computer science term for it - in the current version of the standard (#87), and I've also changed it accordingly in this PR now.

oschulz:
> The get_element indexing is 1-based (right???). The example says [...]

Ah, good point, I didn't speficy that. The question is do we want to use one-based indexing (mathematics/statistic, Julia, MATLAB, Fortran) or zero-based indexing (computer science, C++, Python) in the JSON serialization? I'm sure people may have differing and possibly strong opinions on this, personally I could live with either.

oschulz:
> The from field creates a type-ambiguous resolution problem. Throughout the new accessor functions, from can be "a function or distribution 

But we use the same mechanism for distribution parameters currently, which can also be a reference or literal value, without trouble? I guess the current convention is "a string is always a reference" and that's been fine so far?

> The spec should clarify the resolution order, or better yet, state explicitly that names must be unique across both functions and distributions (which I think is already implied elsewhere, but it's not restated here and it matters a lot for these accessor functions).

Indeed - I think that's always been kind of implied (shared name space across function and distributions) but we should state that very clearly.

oschulz:
> get_elements: The spec says select can be [i1, i2, ...] for 1D or [[i1, j1, ...], [i2, j2, ...], ...] for multi-dimensional. But how does an implementer distinguish between "select elements 1, 2, 3 from a 1D array" ([1, 2, 3]) and "select element at index (1, 2, 3) from a 3D array" ([1, 2, 3])? These are the same JSON.

Yes, that had been on my mind too, that's why I added both `get_element` and `get_elements`. So I think it should be non-ambiguous?

oschulz:
> get_properties: is as optional? If I just want to extract a subset of properties without renaming, do I still need to provide as? The spec doesn't say whether as is required or optional.

I've been wondering about it too - not sure what's best.

oschulz:
> Distribution variates are "always named tuples" but what about scalar distributions? A gaussian_dist has variate x.

Yes, it's something we inherited from RootFit, and more and more I think we do need to let go of it.

oschulz:
> is essentially building a mini functional language for data plumbing which is generally already provided by things like ONNX for you with a well-defined schema... maybe we should be adopting ONNX schemas as subsets of the HS3 spec?

If ONNX provides elements for this that fit in well with what we need, why not? We probably don't want to import more from ONNX than we need, and we may not want to couple the HS3 standard to the ONNX standard and future versions of it - so "pick and clone" instead of reference? But can't hurt to look into it.

oschulz:
Several people who are involved have expressed an interest in having an deeper discussions on HS^3 semantics and related matters before proceeding with incremental changes to the standard text. This might also address issues like "variates have to be records".

I'm leaving this PR open as a draft for the time being, as a point of reference, while we figure these things out.

oschulz:
Depending on the outcome of #87, this PR will need to be adapted, pain points like

```json
	{ 
		"name" : "average_baseline_level", 
		"type" : "get_property", 
		"from" : "avg_bl_dist", 
		"select" : "average_baseline_level" 
	},
```

would go away and other things can likely be simplified/deferred.

oschulz:
Stale, closing.

PR #87, opened 2026-03-06
Make names unique and replace term "named tuple" by "record"

oschulz:
Names should be unique across all variate record entries and all other objects in an HS3 file: I believe this matches RooFit behavior, where pretty much everything is a `RooAbsArg` and names within the `RooAbsArg` namespace must be unique? Except for datasets, which live in `RooAbsData` - but it would limit us to make data point more "special" than variates and parameter points, this asymmetry would be especially annoying in Bayesian contexts. Also, from what I understand it's considered to be bad practice to have duplicate names across `RooAbsArg` and `RooAbsData` anyway, so I think it makes sense to be strict here in HS3.

This also makes awkward future constructions like (proposed in #86)

```json
"distributions":[
	{ 
		"name": "avg_bl_dist", 
		"type": "gaussian_dist", 
		"mean": 0.0, 
		"sigma": 3.0, 
		"x": "average_baseline_level" 
	},
	{ 
		"name": "bl1_dist", 
		"type": "gaussian_dist", 
		"mean": "average_baseline_level", 
		"sigma": "...", 
		"x": "baseline1" 
	},
	...
],
"functions": [ 
	{ 
		"name" : "average_baseline_level", 
		"type" : "get_property", 
		"from" : "avg_bl_dist", 
		"select" : "average_baseline_level" 
	},
	... 
]
```

unnecessary. Instead, this will just become:

```json
"distributions":[
	{ 
		"name": "avg_bl_dist", 
		"type": "gaussian_dist", 
		"mean": 0.0, 
		"sigma": 3.0, 
		"x": "average_baseline_level" 
	},
	{ 
		"name": "bl1_dist", 
		"type": "gaussian_dist", 
		"mean": "average_baseline_level", 
		"sigma": "...", 
		"x": "baseline1" 
	},
	...
]
```

In addition, the clarifies the roles "distribution names" vs. "variate names": Both are global in a common namespace, and the name of the distribution refers to the law of a record of the variate names assigned to it.

Also, "record" is the proper term for "named tuple" in computer science, so it should be used throughout the standard.


oschulz:
Made several changes to clarify semantics further.

cburgard:
Hi @oschulz ,
I don't think this here is true: 
> Names should be unique across all variate record entries and all other objects in an HS3 file: I believe this matches RooFit behavior,

Case in point: Consider a hypothesis testing setup where you have two completely separate models that are used to explain the same data. Both of them will use the same variate, which matches the name of the axis in the data.

I do think that the variate should be unique within one DAC. The opposite (several branches of the graph referring to the same observable) is technically possible, but can IMHO never evaluate to a valid model (unless someone has an example where this might happen).

The critical concept is this I think: within one model, everything that shares the same name (including variates) is conceived of as the same mathematical entity. You might study two different models for the same entity, but never assume two different distributions of the same entity within model.

If I understand your example correctly (which I am not sure I do), you are basically assuming that the mean of one gaussian is the variate of another. This is entirely consistent with the way you would model something in RooFit. A classical example here are the constraint terms. here you would call "average_baseline_level" a global observable.

The critical part for the RooFit side is this: there could be other occurences of "average_baseline_level" in the same workspace. The thing that would really trigger RooFit to tie these two together (in the sense that changes in "average_baseline_level" would propagate correctly everywhere) would be that both of these distributions would be tied into the same likelihood, that is, you would now list
```
"likelihoods" : {
  "distributions" : [ "bl1_dist" ],
  "data" : [ something something x ],
  "aux_distributions" : [ "avg_bl_dist" ]
}
```
To make sure that both of these terms would be picked up by your calculation.

Is that consistent with your proposal, or is this problematic?

Feel free to follow up, I think this is important to get right - even if that means that we need to introduce some additional layer of indirection into the ROOT engine.

oschulz:
~~But aren't all `RooAbsArg` instances in the same flat namespace in a `RooWorkspace` in RooFit?~~

Ah, I think I misunderstood something.

oschulz:
@cburgard I hope it's correct now, could you take another look?

oschulz:
Do we want the rename or should I close this?

PR #96, opened 2026-06-19
Replace ternary operator by ifelse in generic expressions

oschulz:
`ifelse` will be more parser friendly and doesn't carry the C++-implication of "only one branch will be evaluated". HS3 engines should be free to choose if they want to do eager or lazy evaluation.

### DIFF
--- a/docs/chapters/3.1_generic_expressions.md
+++ b/docs/chapters/3.1_generic_expressions.md
@@ -14,7 +14,7 @@ The implementations that support generic expression [must]{.smallcaps} support:
 -   The comparison operators approximately-equal (`x == y`),     exactly-equal (`x === y`), not-approximately-equal (`x != y`),     not-exactly-equal (`x !== y`), less-than (`x < y`), less-or-equal     (`x <= y`), greater-than (`x > y`) and greater-or-equal (`x >= y`). 
 -   The logical operators logical-inverse (`!a`), logical-and     (`a && b`) and logical-or (`a || b`). 
 -   Round brackets to specify the order of operations. 
--   The ternary operator `condition ? result_if_true : result_if_false` 
+-   The conditional expression `ifelse(condition, result_if_true, result_if_false)`
 -   The functions 
     -   `exp(x)`: Euler's number raised to the power of `x` 
     -   `log(x)`: Natural logarithm of `x` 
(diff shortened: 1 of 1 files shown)


cburgard:
I will let some more people comment on this, notably @kratsg and @guitargeek

guitargeek:
I think this is a good suggestion! Since we want to move away from C++-specific syntax, this is a natural evolution.

oschulz:
Are models using this in the wild, or can we just merge this?

oschulz:
@kratsg what's your take on this?

cburgard:
> Are models using this in the wild, or can we just merge this?

I'm pretty sure there are models in the wild that use this, but I would guess that @Phmonski could confirm this.

If this turns out to be true, I suggest that we still mention the "old" syntax, but mark it as deprecated. We can then use existing implementations (notably ROOT) as a possible backend for the upconversion.

Phmonski:
I haven't encountered any models that used "ifelse" statements in their generic expression yet.

cburgard:
> I haven't encountered any models that used "ifelse" statements in their generic expression yet.

ok, in this case, I'm happy with merging this unless @kratsg sees any problems

cburgard:
Outcome from discussion: Make explicit in the text that the behavior w.r.t. callback / lazy evaluation is undefined and may be decided by the engine.

ROOT is also moving away from a pure JIT pipe for the implementation: https://github.com/root-project/root/pull/23209

cburgard:
@oschulz I have added language that clarifies that the order of evaluation is arbitrary. Please review and approve. I will hit merge if you are okay with this.

docs/chapters/3.1_generic_expressions.md
```diff
@@ -37,4 +37,7 @@ Spaces between operators and operands are optional. There must be no space betwe
 Division [must]{.smallcaps} be treated as floating-point division (i.e. `2/3` should be equivalent to `2.0/3.0`). 
 
 The approximately-equal (`a == b`) and the not-approximately-equal operator (`a != b`), [should]{.smallcaps} compare floating-point numbers to within a small multiple of the unit of least precision. 
+
+For the conditional expressions, the order and precedence of evaluation of `condition`, `result_if_true` and `result_if_false` may be freely chosen by the engine.
```
oschulz:
```suggestion
Conditional expression may be evaluated in a lazy fashion (evaluating either  `result_if_true` or `result_if_false` depending on the result of `condition`) or in an eager fashion (evaluating all of  `condition`, `result_if_true`  and `result_if_false` in arbitrary order). Evaluation strategy is left to the engine. This implies that on some engines, large trees of conditional expressions may have a high evaluation cost.
```

oschulz (commented):
ff

PR #97, opened 2026-06-19
standard github reaction emoji

cburgard:
Since our project is growing and more and more people are participating in discussions, I found it increasingly difficult to communicate precisely without cluttering the discussion thread with comments that do not actually contribute to the merit of the argument being made.
Also, many of the standard github reaction emoji only map very inaccurately to the types of discussions we are having here (standardization, consensus-finding, community feedback).

Hence, I drafted this PR as an attempt to codify the standard allowed github emoji reactions in a way that make them useful for our project. 

You might find this silly, and maybe it is, but I still think it can be useful in streamlining our discussions.

Please let me know what you think about this enterprise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,22 @@ proposals (draft or pending) folder and create a PR to discussion them.
 If you are interest to become part of the core committee, please open an
 issue. Anyone is allowed to join.
 
+## Emoji-Reactions to Issues, PRs, Discussions and Comments
+
+To keep thread notifications low and avoid comment spam, we use GitHub reactions as a structured tool for project governance. We invite everyone to use these specific emojis. 
+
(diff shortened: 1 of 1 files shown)


stalbrec:
Do we want to distinguish between 👍 and +1/upvote in Github Discussions or discourage people to use upvotes altogether?

cburgard:
> Do we want to distinguish between 👍 and +1/upvote in Github Discussions or discourage people to use upvotes altogether?

I think :+1:  is the quintessential upvote, so i would treat them as equal - or do you have a different take here?

stalbrec:
Semantically, I'd agree.  
But if we ever needed voting where provenance matters (i.e. seeing who cast each vote), we'd have to use the :+1: reaction rather than the upvote button, since GitHub doesn't expose upvote provenance. 
However, I think we should clarify this explicitly for these not-every-day use cases, rather than adding this here.

PR #101, opened 2026-07-01
Introduce factorized normalization for `product_dist`

cburgard:
This PR extends `product_dist` with an optional `normalization` field that allows the normalization of a product distribution to be specified explicitly.

By default, a product distribution is normalized globally over all observables. The new `normalization` field instead allows subsets of factors to be normalized independently over specified observables. The resulting normalization factor is therefore no longer required to be a constant; it may depend on observables that are not integrated within a particular normalization group.

This provides a natural and implementation-independent way to represent conditional probability densities without introducing a dedicated `conditional_dist` object. Instead, conditionality is expressed as a property of the normalization structure of the product rather than as an intrinsic property of the individual distributions. For example, a factorization such as

$$
p(x,y,z) = p(x \mid y), p(y,z), p(z)
$$

can be represented as a product of ordinary distributions together with a normalization group that normalizes the first factor over `x`.

The proposal also defines the necessary constraints on normalization groups to ensure that the resulting normalization is well-defined and unambiguous.

### DIFF
--- a/docs/chapters/1_introduction.md
+++ b/docs/chapters/1_introduction.md
@@ -6,7 +6,9 @@ bibliography: hs3.bib
 *This section is non-normative.* 
 
 It is widely agreed upon that the publication of likelihood models from high energy physics experiments is imperative for the preservation and public access to these results. Detailed arguments have been made elsewhere already [@Cranmer_2022]. This document sets out to provide a standardized format to publish and archive statistical models of any size and across a wide range of mathematical formalisms in a way that is readable by both humans and machines. 
+
 With the introduction of `pyhf` [@pyhf; @pyhf-joss], a `JSON` format for likelihood serialization has been put forward. However, an interoperable format that encompasses likelihoods with a scope beyond stacks of binned histograms was lacking. With the release of `ROOT` 6.26/00 [@root] and the experimental `RooJSONFactoryWSTool` therein, this gap has now been filled. 
+
 This document sets out to document the syntax and features of the Statistics Serialization Standard (HS${}^3$) for likelihoods and statistical models in general, as to be adopted by any HS${}^3$-compatible statistics framework. The examples in this document employ the `JSON` notation, but are intended to encompass also representations as `YAML` or `TOML`. 
 
 ## How to use this document 
@@ -15,14 +17,30 @@ Please note that this document as well as the HS${}^3$ standard are still in dev
 
 ## Statistical semantics of HS${}^3$ {#sec:hs3-semantics} 
 
+A **variable** is a mathematical quantity whose value is chosen or generated,
+rather than directly computed from other quantities. A **parameter** is a
+variable whose value is chosen by a user of the model, such as a minimizer.
+There is no expectation that a distribution is normalized with respect to
+its parameters, or that parameters themselves have specified statistical
+properties or a distribution from which they can be drawn. A **variate**,
+also called an **observable**, is a statistical quantity modeled by a
+distribution; these two terms are used interchangeably in HS${}^3$.
+Both parameters and variates are variables. Their roles are relative to a
+particular use: the same variable may be a parameter in one use and a
+variate in another, for example a parameter of a likelihood model and a
+variate of a prior distribution.
+
 ### Statistical models, probability distributions and parameters {#sec:models-and-parameters} 
+
 HS${}^3$ takes a "forward-modelling" approach throughout: a statistical model $m$ maps a space $\Theta$ of free parameters $\theta$ to a space of probability distributions that describe the possible outcomes of a specific experiment. For any given parameters $\theta$, the model returns a concrete probability distribution $m(\theta)$. Observed data $x$ is then treated as random variate, presumed to be drawn from the model: $x \sim m(\theta)$. 
 Parameters in HS${}^3$ are always named, so semantically the elements of a parameter space $\Theta$ are records $\theta$ (i.e. tuples in which every entry has a name). Even if there is only a single free parameter, $\theta$ should be thought of as a single-entry record. In the current version of this standard, parameter tuples must be flat and only consist of real numbers, vector-valued or nested entries are not supported yet. Future versions of the standard will likely be less restrictive, especially with respect to discrete or vector-valued parameters. Parameter domains defined as part of this standard may be of various types, even though the current version only supports product domains. Future versions are likely to be less restrictive in this regard. 
 Mathematically and computationally, it is often convenient to treat parameters as flat real-valued vectors instead of records, it is the responsibility of the implementation to map between these different views of parameters when and where necessary.
+
 Probability distributions in HS${}^3$ (see 
 [Distributions](#sec:distributions){reference-type="ref" reference="sec:distributions"}) are typically parameterized. Any instance of a distribution that has some of its parameters bound to names instead of concrete values, e.g. $m = d_{\mu = a, \sigma = 0.7, \lambda = b, ...}$, constitutes a valid statistical model $m(\theta)$ with model parameters $\theta = (a, b)$. 
 When probability distributions are combined via a Cartesian product (see 
 [Product distribution](#sec:product-distribution){reference-type="ref" reference="sec:product-distribution"}), then the records that contain their free parameter values are concatenated. So $m = d_{\mu = a, \sigma = b} \times d_{\mu = c, \sigma = 0.7}$ constitutes a model $m(\theta)$ with model parameters $\theta = (a, b, c)$.
+
 Distribution parameters may also be bound to the output of functions, and if those functions have inputs that are bound to names instead of values (or again bound to such functions, recursively), then those names become part of the model parameters. A configuration $m = d_{\mu = a, \sigma = 0.7, \lambda = f}, f = \texttt{sum}(4.2, g), g = \texttt{sum}(1.3, b)$, for example, also constitutes a (different) model $m(\theta)$ with parameters $\theta = (a, b, c)$. 
 If all parameter values of a probability distribution are set to concrete values, so if there are not directly (or indirectly via functions) bound to names, we call this probability distribution a concrete distribution here. Such distributions can be used as Bayesian 
 priors (see [Bayesian inference](#sec:bayesian-inference){reference-type="ref" reference="sec:bayesian-inference"}). 
@@ -40,8 +58,10 @@ multiple HS${}^3$ documents together, e.g. for combined analyses.
 
 ### Probability density functions (PDFs) {#sec:what-is-a-pdf} 
 Statistics literature often discriminates between probability density functions (PDF) for continuous probability distributions and probability mass functions (PMF) for discrete probability distributions. This standard use the term PDF for both continuous and discrete distributions. The concept of density is to be understood in terms of densities in the realm of measure theory here, that is the density of a probability measure (distribution) is its Radon-Nikodym derivative in respect to an (implied) reference measure. 
+
 The choice of reference measure would be arbitrary in principle, which scales likelihood functions 
 (Sec. [Likelihood](#sec:likelihood-definition){reference-type="ref" reference="sec:likelihood-definition"}) by a constant factor that depends on choice of reference. In this standard, a specific reference measures is implied for each probability distribution, typically the Lebesgue measure for continuous distributions and the counting measure for discrete distributions. The standard aims to match the PDF (resp.&nbsp;PMF) most commonly used in literature for each specific probability distribution and the mathematical form of the PDF is documented explicitly for each distribution in the standard. So within HS${}^3$, probability densities and likelihood functions are unambiguous. 
+
 Here we use $\text{PDF}(m(\theta), x)$ to denote the density value of the probability distribution/measure $m$, parameterized by $\theta$, at the point/variate $x$, in respect to the implied reference for $m$. 
 
 ### Observed and simulated data {#sec:data-generation} 
--- a/docs/chapters/2.1.2_composite_distributions.md
+++ b/docs/chapters/2.1.2_composite_distributions.md
@@ -121,8 +121,9 @@ This can be represented by the product
 ```
 
 Here, $p_1$ is normalized independently over `x`, while the remaining subproduct $p_2 p_3$ is normalized over the remaining observables ($y$ and $z$), so
+
 $$
-\ensuremath{\mathcal{M}}&= \int dx p_1(x, y) \int\int dy dz p_2(y, z) p_3(z)
+\begin{aligned} \ensuremath{\mathcal{M}}&= \int dx p_1(x, y) \int\int dy dz p_2(y, z) p_3(z) \end{aligned} 
 $$
 
 
--- a/docs/chapters/2.1.2_composite_distributions.md
+++ b/docs/chapters/2.1.2_composite_distributions.md
@@ -66,20 +66,82 @@ If all coefficients are given, their sum is used as the expected number of event
 
 #### Product distribution {#sec:product-distribution label="sec:product-distribution"} 
 
-The product of independent distributions $f_i$ is defined as 
-
-
+The product of distributions $f_i$ is defined as 
 
 $$
-\begin{aligned} \text{ProductPdf}(x) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f(x). \end{aligned} 
+\begin{aligned} \text{ProductPdf}(x_0, x_1, \dots) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f_i(x_0, x_1, \dots). \end{aligned} 
 $$
 
 
 
--   `name`: custom string 
--   `type`: `product_dist` 
--   `x`: name of the variable $x$ ([optional]{.smallcaps}, since the     variable is fully defined by the factors) 
--   `factors`: array of names referencing distributions 
+-`name`: custom string 
+-`type`: `product_dist` 
+-`x`: array of names of the variables $x$ ([optional]{.smallcaps}, since the variable is fully defined by the factors) 
+-`factors`: array of names referencing distributions 
+-`normalization`: [optional]{.smallcaps} possibility to define the normalization $\ensuremath{\mathcal{M}}$ of the product (see below)
+
+
+A product distribution is by default, interpreted as the product of its factors with a single global normalization over all observables, that is,
+
+$$
+\ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_0, x_1, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_0, x_1, \dots)
+$$
+
+Here, $d\Omega$ denotes the product measure over all observables in the current analysis context.
+
+The normalization factor $\ensuremath{\mathcal{M}}$ denotes the
+denominator applied to the product distribution.  The optional
+([optional]{.smallcaps} `normalization` field specifies the
+normalization of subproducts by declaring groups of factors that are
+independently normalized over specified observables.  Each entry in
+`normalization` identifies
+
+* a set of factors forming a subproduct, and
+* the observables over which that subproduct is individually normalized.
+
+After all explicitly declared normalization groups have been applied, the remaining factors are multiplied together and normalized over every observable that has not already been assigned to a normalization group.
+When the [optional]{.smallcaps}  `normalization` is used,  $\ensuremath{\mathcal{M}}$ is, in general, no longer a constant but may depend on observables that are not integrated within one or more normalization groups.
+
+This mechanism expresses **conditional probability densities** through the normalization structure of the product, rather than by assigning conditionality to individual distributions.
+
+For example, consider the factorization
+
+$$
+p(x,y,z) = p_1(x,y) p_2(y,z) p_3(z)
+$$
+
+This can be represented by the product
+
+```json
+"type": "product_dist"
+"factors": ["P1", "P2", "P3"]
+"normalization": [
+	{ "factors": [P1], "over": ["x"] }
+]
+```
+
+Here, $p_1$ is normalized independently over `x`, while the remaining subproduct $p_2 p_3$ is normalized over the remaining observables ($y$ and $z$), so
+$$
+\ensuremath{\mathcal{M}}&= \int dx p_1(x, y) \int\int dy dz p_2(y, z) p_3(z)
+$$
+
+
+Note that  $\ensuremath{\mathcal{M}} = \ensuremath{\mathcal{M}}(y)$  is now a function of the observable $y$, which is not integrated during the normalization of the first subproduct. 
+This normalization structure is mathematically equivalent to the conditional factorization
+
+$$
+p(x,y,z) = p_1(x \mid y)\, p_2(y,z)\, p_3(z),
+$$
+
+where the conditional density $p_1(x \mid y)$ is induced by normalizing the first subproduct over $x$. This avoids assigning conditionality as an intrinsic property of the individual distributions, instead expressing it through the normalization structure of the product.
+
+The following constraints apply:
+
+* Each factor [shall]{.smallcaps} appear in at most one `normalization` entry.
+* An observable  [shall]{.smallcaps} be normalized by at most one factorized subproduct.
+* If an observable is normalized by a factorized subproduct, every factor depending on that observable  [shall]{.smallcaps} belong to that same subproduct.
+* Factors and observables not referenced by any entry participate in the implicit global normalization.
+
 
 #### Efficiency product distribution
 
--- a/docs/chapters/2.1.2_composite_distributions.md
+++ b/docs/chapters/2.1.2_composite_distributions.md
@@ -84,7 +84,7 @@ $$
 A product distribution is by default, interpreted as the product of its factors with a single global normalization over all observables, that is,
 
 $$
-\ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_0, x_1, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_0, x_1, \dots)
+\begin{aligned} \ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_0, x_1, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_0, x_1, \dots) \end{aligned} 
 $$
 
 Here, $d\Omega$ denotes the product measure over all observables in the current analysis context.
--- a/docs/chapters/2.1.2_composite_distributions.md
+++ b/docs/chapters/2.1.2_composite_distributions.md
@@ -69,7 +69,7 @@ If all coefficients are given, their sum is used as the expected number of event
 The product of distributions $f_i$ is defined as 
 
 $$
-\begin{aligned} \text{ProductPdf}(x_0, x_1, \dots) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f_i(x_0, x_1, \dots). \end{aligned} 
+\begin{aligned} \text{ProductPdf}(x_1, x_2, \dots) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f_i(x_1, x_2, \dots). \end{aligned} 
 $$
 
 
@@ -84,7 +84,7 @@ $$
 A product distribution is by default, interpreted as the product of its factors with a single global normalization over all observables, that is,
 
 $$
-\begin{aligned} \ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_0, x_1, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_0, x_1, \dots) \end{aligned} 
+\begin{aligned} \ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_1, x_2, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_1, x_2, \dots) \end{aligned} 
 $$
 
 Here, $d\Omega$ denotes the product measure over all observables in the current analysis context.
--- a/docs/chapters/2.1.2_composite_distributions.md
+++ b/docs/chapters/2.1.2_composite_distributions.md
@@ -104,10 +104,10 @@ When the [optional]{.smallcaps}  `normalization` is used,  $\ensuremath{\mathcal
 
 This mechanism expresses **conditional probability densities** through the normalization structure of the product, rather than by assigning conditionality to individual distributions.
 
-For example, consider the factorization
+For example, consider the unnormalized factorization
 
 $$
-p(x,y,z) = p_1(x,y) p_2(y,z) p_3(z)
+p(x,y,z) \propto p_1(x,y) p_2(y,z) p_3(z).
 $$
 
 This can be represented by the product
@@ -120,21 +120,45 @@ This can be represented by the product
 ]
 ```
 
-Here, $p_1$ is normalized independently over `x`, while the remaining subproduct $p_2 p_3$ is normalized over the remaining observables ($y$ and $z$), so
+Here, $p_1$ is normalized independently over `x`, while the remaining
+subproduct $p_2 p_3$ is normalized over the remaining observables
+($y$ and $z$). The factorized normalization denominator is therefore
 
 $$
-\begin{aligned} \ensuremath{\mathcal{M}}&= \int dx p_1(x, y) \int\int dy dz p_2(y, z) p_3(z) \end{aligned} 
+\begin{aligned}
+\ensuremath{\mathcal{M}}(y)
+&=
+\left[\int dx\,p_1(x,y)\right]
+\left[\int\!\!\int dy\,dz\,p_2(y,z)p_3(z)\right].
+\end{aligned}
 $$
 
-
-Note that  $\ensuremath{\mathcal{M}} = \ensuremath{\mathcal{M}}(y)$  is now a function of the observable $y$, which is not integrated during the normalization of the first subproduct. 
-This normalization structure is mathematically equivalent to the conditional factorization
+Note that $\ensuremath{\mathcal{M}}=\ensuremath{\mathcal{M}}(y)$ is now
+a function of $y$, which is not integrated during the normalization of
+the first subproduct. The distribution encoded by this normalization
+structure is
 
 $$
-p(x,y,z) = p_1(x \mid y)\, p_2(y,z)\, p_3(z),
+\begin{aligned}
+p(x,y,z)
+&=
+\frac{p_1(x,y)}
+     {\int dx\,p_1(x,y)}
+\,
+\frac{p_2(y,z)p_3(z)}
+     {\int\!\!\int dy\,dz\,p_2(y,z)p_3(z)}
+\\
+&=
+p_1(x\mid y)\,p(y,z).
+\end{aligned}
 $$
 
-where the conditional density $p_1(x \mid y)$ is induced by normalizing the first subproduct over $x$. This avoids assigning conditionality as an intrinsic property of the individual distributions, instead expressing it through the normalization structure of the product.
+Here, the conditional density $p_1(x\mid y)$ is induced by normalizing
+the first subproduct over $x$, and $p(y,z)$ is the remaining subproduct
+$p_2(y,z)p_3(z)$ normalized over $y$ and $z$. This is the chain-rule
+factorization of the normalized joint distribution. It avoids assigning
+conditionality as an intrinsic property of an individual distribution,
+instead expressing it through the normalization structure of the product.
 
 The following constraints apply:
 
--- a/docs/chapters/2.1.2_composite_distributions.md
+++ b/docs/chapters/2.1.2_composite_distributions.md
@@ -66,43 +66,50 @@ If all coefficients are given, their sum is used as the expected number of event
 
 #### Product distribution {#sec:product-distribution label="sec:product-distribution"} 
 
-The product of distributions $f_i$ is defined as 
+The product of independent distributions $f_i$ is defined as
 
 $$
-\begin{aligned} \text{ProductPdf}(x_1, x_2, \dots) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f_i(x_1, x_2, \dots). \end{aligned} 
+\begin{aligned} \text{ProductPdf}(x) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f_i(x). \end{aligned}
 $$
 
+-   `name`: custom string
+-   `type`: `product_dist`
+-   `x`: name of the variable $x$ ([optional]{.smallcaps}, since the variable is fully defined by the factors)
+-   `factors`: array of names referencing distributions
+
+The variates of the factors [must]{.smallcaps} be pairwise disjoint. Consequently, the normalization factorizes into the product of the normalizations of the factors. This structure permits efficient normalization and sampling by treating every factor independently.
 
 
--`name`: custom string 
--`type`: `product_dist` 
--`x`: array of names of the variables $x$ ([optional]{.smallcaps}, since the variable is fully defined by the factors) 
--`factors`: array of names referencing distributions 
--`normalization`: [optional]{.smallcaps} possibility to define the normalization $\ensuremath{\mathcal{M}}$ of the product (see below)
+#### Point-product distribution {#sec:point-product-distribution label="sec:point-product-distribution"}
 
-
-A product distribution is by default, interpreted as the product of its factors with a single global normalization over all observables, that is,
+A point-product distribution is the normalized pointwise product of distributions whose variates may overlap:
 
 $$
-\begin{aligned} \ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_1, x_2, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_1, x_2, \dots) \end{aligned} 
+\begin{aligned}
+\text{PointProductPdf}(x_1, x_2, \dots)
+&= \frac{1}{\ensuremath{\mathcal{M}}}
+   \prod_{i=1}^{n} f_i(x_1, x_2, \dots).
+\end{aligned}
 $$
 
-Here, $d\Omega$ denotes the product measure over all observables in the current analysis context (see Section [Analyses](#sec:analyses)).
+-   `name`: custom string
+-   `type`: `point_product_dist`
+-   `x`: array of names of the variables $x$ ([optional]{.smallcaps}, since the variables are fully defined by the factors)
+-   `factors`: array of names referencing distributions
+-   `conditionals`: [optional]{.smallcaps} array declaring conditional subproducts (see below)
 
-The normalization factor $\ensuremath{\mathcal{M}}$ denotes the
-denominator applied to the product distribution.  The optional
-([optional]{.smallcaps} `normalization` field specifies the
-normalization of subproducts by declaring groups of factors that are
-independently normalized over specified observables.  Each entry in
-`normalization` identifies
+Without `conditionals`, the product has a single global normalization over all its variates. In an analysis context, only variables interpreted as observables are integrated; variables interpreted as parameters are held fixed (see Section [Analyses](#sec:analyses)). Thus,
 
-* a set of factors forming a subproduct, and
-* the observables over which that subproduct is individually normalized.
+$$
+\begin{aligned}
+\ensuremath{\mathcal{M}}
+&= \int d\Omega\,\prod_{i=1}^{n} f_i(x_1, x_2, \dots),
+\end{aligned}
+$$
 
-After all explicitly declared normalization groups have been applied, the remaining factors are multiplied together and normalized over every observable that has not already been assigned to a normalization group.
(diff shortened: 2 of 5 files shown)


cburgard:
Hi @oschulz @BJMCox , this one is addressing your request for conditional pdfs.
Comments / improvement suggestions are highly welcome!

oschulz:
Does this approach have technical advantages with respect to RooFit import/export or so?

I think it might be better to express conditionality more directly. This PR implements a very density-centric view, which we have avoided in other parts of the spec. Maybe a `dependent_product_dist` would be better? Dependent and independent products are fundamentally different (and may to very different code structures), so it seems a bit forced to select between them with an option.

cburgard:
> Does this approach have technical advantages with respect to RooFit import/export or so?
> 
> I think it might be better to express conditionality more directly. This PR implements a very density-centric view, which we have avoided in other parts of the spec. Maybe a `dependent_product_dist` would be better? Dependent and independent products are fundamentally different (and may to very different code structures), so it seems a bit forced to select between them with an option.

Hi @oschulz ,

Thanks for sharing your opinion.

There are several design goals I tried to achieve with this approach.

The main one was this: In the spec, we have so far refrained from differentiating between observables and parameters on the level of an individual distribution, that is: A distribution is generally not allowed to assume which if its variables have the role of parameter and which of them have the role of observable. That is only fixed by the interpretation. 
So the main goal here was to find a way to conceptualize conditional observables without prescribing that they are, in fact, observables. The crucial "insight" here is this:
The distributions will generally know which of the variables are "allowed" to be observables, but the model can always "demote" (so to speak) them to parameters. 
The only real difference between the mathematics as far as conditional and unconditional observables are concerned is how the integrals are performed, so I tried to come up with a syntax that focuses on prescribing the integration rules of the product while attempting to invoid imbuing it with a statistical interpretation already.
From my side, thus, there is no particular reason why the concept of density needs to be front-and-center of this syntax - please feel free to suggest a different way! - but what was a conscious choice was to focus on the integration rules rather than the interpretation.

Not-so-coincidentally, the RooFit implementation also only distinguishes between the conditional and unconditional case by prescribing integration rules, but in ways that are a bit more limiting than the syntax I have proposed here.

As for your concrete proposal: What would be the difference between the `dependent_product_dist` and the regular `product_dist`? I am not against distinguishing it into two, but if the only real difference would be that the dependent version would make specifiying the integration groups mandatory while the regular one would not allow it, I'm not sure that's a worthwhile distinction - but others may disagree here, and I'm happy to change the proposal accordingly if a distinctive second distribution type is what makes people happy here.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,83 @@ If all coefficients are given, their sum is used as the expected number of event
...
+p(x,y,z) = p_1(x,y) p_2(y,z) p_3(z)
+$$
+
+This can be represented by the product
+
+```json
+"type": "product_dist"
+"factors": ["P1", "P2", "P3"]
+"normalization": [
+	{ "factors": [P1], "over": ["x"] }
```
BJMCox: Should this not be "P1"?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,83 @@ If all coefficients are given, their sum is used as the expected number of event
...
+
+
+A product distribution is by default, interpreted as the product of its factors with a single global normalization over all observables, that is,
 
 $$
-\begin{aligned} \text{ProductPdf}(x) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f(x). \end{aligned} 
+\begin{aligned} \ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_0, x_1, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_0, x_1, \dots) \end{aligned} 
 $$
 
+Here, $d\Omega$ denotes the product measure over all observables in the current analysis context.
```
BJMCox: As far as I understand the spec avoids a fixed split between observables and parameters. I think that we can phrase this in terms of variables and domains (like we do with distributions). Also, I do not think that we have defined what "current analysis context" means

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,83 @@ If all coefficients are given, their sum is used as the expected number of event
...
+-`x`: array of names of the variables $x$ ([optional]{.smallcaps}, since the variable is fully defined by the factors) 
+-`factors`: array of names referencing distributions 
+-`normalization`: [optional]{.smallcaps} possibility to define the normalization $\ensuremath{\mathcal{M}}$ of the product (see below)
+
+
+A product distribution is by default, interpreted as the product of its factors with a single global normalization over all observables, that is,
 
 $$
-\begin{aligned} \text{ProductPdf}(x) &= \frac{1}{\ensuremath{\mathcal{M}}}\prod_i^n f(x). \end{aligned} 
+\begin{aligned} \ensuremath{\mathcal{M}}&= \int d\Omega \prod_i^n f_i(x_0, x_1, \dots) = \int\int\dots dx_1 dx_2 \dots \prod_i^n f_i(x_0, x_1, \dots) \end{aligned}
```
BJMCox: Measure starts at dx_1, but this starts at x_1. Should start at x_1 or measure should start at dx_0 I think

docs/chapters/2.1.2_composite_distributions.md
cburgard: absolutely, thanks for catching this!

docs/chapters/2.1.2_composite_distributions.md
cburgard: > As far as I understand the spec avoids a fixed split between observables and parameters.

Yes, that is correct - at least as far as the defining features of any one distribution are concerned.
The reason why I deemed the distinction "okay" here is precisely because the restriction "in the current analysis context", which you take offense with, so let me try to elaborate.
The role of any one variable (observable or parameter) is only defined in the context of a specific operation to be conducted (e.g. building a likelihood with a dataset - then every variable that is contained in the dataset is an observable. or sampling from a distribution - then every variable that is requested to be part of the generated dataset is an observable. everything else is a parameter). 
And because the "analysis" object refers to a "likelihood", which contains "distributions" and "data", where the pairing of distributions to data then defines the roles of the variables (observables or parameters), the formulation "in the current analysis context" is meant to reflect this precise relation.
However, I see that this is not at all explicit enough, and likely confusing. Do you have a recommendation of how this could be rephrased?

docs/chapters/2.1.2_composite_distributions.md
cburgard: Yes for sure, thanks for catching this!

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,83 @@ If all coefficients are given, their sum is used as the expected number of event
...
+$$
+\begin{aligned} \ensuremath{\mathcal{M}}&= \int dx p_1(x, y) \int\int dy dz p_2(y, z) p_3(z) \end{aligned} 
+$$
+
+
+Note that  $\ensuremath{\mathcal{M}} = \ensuremath{\mathcal{M}}(y)$  is now a function of the observable $y$, which is not integrated during the normalization of the first subproduct. 
+This normalization structure is mathematically equivalent to the conditional factorization
+
+$$
+p(x,y,z) = p_1(x \mid y)\, p_2(y,z)\, p_3(z),
```
BJMCox: Line 110 says $p(x,y,z) = p_1(x,y)\, p_2(y,z)\, p_3(z)$.
Line 134 says $p(x,y,z) = p_1(x \mid y)\, p_2(y,z)\, p_3(z)$.

If both hold, dividing one by the other gives $\int dx\, p_1(x,y) = 1$ for every $y$, i.e. $p_1$ was already normalized in $x$.
In that case the normalization group does nothing, and line 130 (that $\mathcal{M}$ is a function of $y$) is false.

Line 134 has another problem as well I think as it drops the normalization of the remaining subproduct. 

Line 126 says $\mathcal{M} = \int dx\, p_1(x,y) \iint dy\, dz\, p_2(y,z)\, p_3(z)$, so the serialized object produces

$$\frac{p_1\, p_2\, p_3}{\mathcal{M}} = p_1(x \mid y)\, \frac{p_2(y,z)\, p_3(z)}{\iint dy\, dz\, p_2\, p_3},$$

while the right-hand side of line 134 integrates to $\iint dy\, dz\, p_2\, p_3$, not 1.

I think this can be fixed by the following:
- Line 110: write $p(x,y,z) \propto p_1(x,y)\, p_2(y,z)\, p_3(z)$.
- Line 134: write the encoded distribution as above, noting that the second fraction is the remaining subproduct $p_2 p_3$ normalized over $y$ and $z$.

This is the chain-rule form $p(x,y,z) = p(x \mid y)\, p(y,z)$, and it makes line 130 correct as $\mathcal{M}$ inherits the $y$-dependence of $\int dx\, p_1(x,y)$.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,83 @@ If all coefficients are given, their sum is used as the expected number of event
...
+$$
+
+where the conditional density $p_1(x \mid y)$ is induced by normalizing the first subproduct over $x$. This avoids assigning conditionality as an intrinsic property of the individual distributions, instead expressing it through the normalization structure of the product.
+
+The following constraints apply:
+
+* Each factor [shall]{.smallcaps} appear in at most one `normalization` entry.
+* An observable  [shall]{.smallcaps} be normalized by at most one factorized subproduct.
+* If an observable is normalized by a factorized subproduct, every factor depending on that observable  [shall]{.smallcaps} belong to that same subproduct.
+* Factors and observables not referenced by any entry participate in the implicit global normalization.
```
BJMCox, line 171: Also need:
- Any entry's factors must be a subset of the product's factors
- Any given 'over' observable must be a variable of at least one factor in its group (as otherwise the normalisation integral for that group is incorrect as it will gain a bare volume factor)

docs/chapters/2.1.2_composite_distributions.md
BJMCox: In honesty I am in general in favour of more distinction between "observables" and "parameters". I was more pointing this out as this is new here to my knowledge.

I am not 100% sure on how I might improve the wording here - it is rather a delicate and implicit construct. I think that, if we define "current analysis context" explicitly as you have above (perhaps somewhat in a more formulaic manner) then this would be easier to parse?

docs/chapters/2.1.2_composite_distributions.md
cburgard: You are absolutely correct, please take a look at the "fixed" version I have committed and see if it correctly addresses your point.

docs/chapters/2.1.2_composite_distributions.md
cburgard: good catch, thanks!

BJMCox:
My thoughts on the philosophy of this PR: I do not believe the PR to be *overly* density centric. From what I understand the spec does not fix the role of an observable or a parameter at the level of a distribution, and this PR follows that by prescribing integration rules (rather than a particular statistical interpretation). I presume that RooFit operates in a similar manner.

From what I understand, a `conditional_dist` construct would require to fix variable roles (observable / parameter), which we have not yet done in the spec. This PR has every conditional factor be a regular distribution and stores the conditional structure in another construct, which is more consistent with the rest of the spec.

However, there are issues with the proposed prose. The primary one is that many lines speak of observables, which fixes the variable roles which this proposal seems geared to not fix. I think this can easily be addressed by rephrasing in terms of variables.

I am not sure that this proposal is any more density centric than the rest of the spec. For example, in the distributions chapter we generally speak in terms of densities (e.g. referencing integrals over domains). However, in other parts such as the introduction and when talking of the Poisson point process we talk in terms of measure theory. I would not dismiss that this PR is density centric, however in the context it lies it is not an outlier.

I would agree that switching between dependent and independent products is a risk, especially if an implementation is permissive in how it reads it may not support one method and silently default to the other. We have discussed this before however, and having it be normative that implementations do not make non-equivalent choices silently would resolve this.

Finally, I note that the proposed constraints operate only over observables, and something being an observable is fixed by the analysis context. Therefore, it is potentially possible for a product distribution to be valid under one likelihood and ill formed under another. I am not completely sure how this could be addressed - although I imagine most analyses would be fine.

docs/chapters/2.1.2_composite_distributions.md
BJMCox: I believe the new formulation to be correct

docs/chapters/2.1.2_composite_distributions.md
cburgard: I gave it a shot, please take a look here. Does this make sense to you?

https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/pull/101#discussion_r3713084466

docs/chapters/2.7_analyses.md
```diff
@@ -29,3 +29,18 @@ All parameters of all distributions in the likelihood must either be listed unde
   ... 
 ]
 ``` 
+
+An analysis defines the notion of an **analysis context** in which a
```
cburgard, line 33: @BJMCox

cburgard:
> My thoughts on the philosophy of this PR: I do not believe the PR to be _overly_ density centric. From what I understand the spec does not fix the role of an observable or a parameter at the level of a distribution, and this PR follows that by prescribing integration rules (rather than a particular statistical interpretation). I presume that RooFit operates in a similar manner.
> 
> From what I understand, a `conditional_dist` construct would require to fix variable roles (observable / parameter), which we have not yet done in the spec. This PR has every conditional factor be a regular distribution and stores the conditional structure in another construct, which is more consistent with the rest of the spec.
> 
> However, there are issues with the proposed prose. The primary one is that many lines speak of observables, which fixes the variable roles which this proposal seems geared to not fix. I think this can easily be addressed by rephrasing in terms of variables.
> 
> I am not sure that this proposal is any more density centric than the rest of the spec. For example, in the distributions chapter we generally speak in terms of densities (e.g. referencing integrals over domains). However, in other parts such as the introduction and when talking of the Poisson point process we talk in terms of measure theory. I would not dismiss that this PR is density centric, however in the context it lies it is not an outlier.
> 
> I would agree that switching between dependent and independent products is a risk, especially if an implementation is permissive in how it reads it may not support one method and silently default to the other. We have discussed this before however, and having it be normative that implementations do not make non-equivalent choices silently would resolve this.
> 
> Finally, I note that the proposed constraints operate only over observables, and something being an observable is fixed by the analysis context. Therefore, it is potentially possible for a product distribution to be valid under one likelihood and ill formed under another. I am not completely sure how this could be addressed - although I imagine most analyses would be fine.

Thanks a lot for your detailed review, which is greatly appreciated!

You are exactly right in that this PR tries to walk the (thin?) line required to define conditional products without immersing itself into an inherent distinction between observables and parameters. If you feel like the language needs to be cleaned up to this extent, I'm happy to take care if that if you can point me to the concrete instances you find problematic.

The fact that this allows the same distribution to be referenced in multiple analyses, with potential differences between how variables are interpreted as observables or parameters between them. I hope (?) that the formulation I have chosen here should be "safe" in the sense that if any variable is referenced in the normalization group section, it should be treated correctly if that variable is an observable, and ignored (=equivalent to the corresponding part being removed) if that variable is a parameter. If you think this is not the case, please let me know so we can adjust it accordingly. Ideally, the validity of a concrete realization of the spec should not depend on the analysis context.

As to your general point on how density-centric and/or measure-theory-focused the language in the spec is, it is true that there is some variation in how each individual PR is formulated. I would not be opposed in taking a more measure-theory-focused approach to wording things, but would ask someone more well-versed to suggest the corresponding wordings at the appropriate places (as long as we can agree that the representation itself is fine).
The alternative to this would be that we merge this PR despite the density-centric language, and raise a more general issue to harmonize the wording across the standard w.r.t. the measure-theory vs. density distinction. Which version do you prefer? (also, @oschulz)

I completely agree with this point here:
> We have discussed this before however, and having it be normative that implementations do not make non-equivalent choices silently would resolve this.

If you have a concrete suggestion on where/how we should define this, I'm happy to adopt it.

docs/chapters/2.1.2_composite_distributions.md
BJMCox: The proposed description is clear to me 😄

docs/chapters/2.7_analyses.md
BJMCox: On reading this is clear to me. I would not be opposed to a few more eyes on it, however

BJMCox:
I think that guidance / norms for implementations might require another (sub)section, as I cannot find anywhere that it neatly fits. It should not be something that general users need to read or worry about I would not think. 

Given the adjustments made to the normalisation due to feedback, this *should* be safe now (w.r.t. to variable interpretation across analyses)- I will need to crack out pen and paper to check for sure though.

I have no preference regarding density over measure theory, or vice-versa. Density notation is more convenient for most readers, however I think that the measure theoretic notation is closer to how RooFit operates (although I am unsure there, this is based on 2nd hand understanding). It would be more convenient to normalise on one or the other - my thinking is that this should be the decision of the main spec authors, as they would be the ones writing the majority of it. Having things be wrong because we standardised on an unfamiliar notation is very counterproductive.

mmikhasenko:
notes from the discussion:
- `p1(x,y) * p2(y,x) * p3(y)` as product of densities (~`point_product`).
- `p1(x) * p2(y)` is a "standard" product, that has a efficient normalization and sampling algorithm.
- conditional `p(x|y)`, and two-var. distributions, `p(x,y)` are different objects 

- [roofit doc](https://root.cern/doc/master/classRooProdPdf.html#ab70833bd3f4642198e8f8b3199497f76) on `RooProdPdf` misses one integral

 
Screenshot 2026-09-04 at 14 54 09

cburgard:
proposed way forward:

- product_dist stays the way it is currently in the standard and is explicitly for orthogonal variates
- this proposal will be renamed into a point_product_dist and will retain the behavior lined out here, but will switch from using the vocabulary of normalizations to an explicit vocabulary of conditionality

cburgard:
My latest commit implements the proposals from todays meeting. @oschulz @szabo137 please let me know if this is in line with what you were hoping for!

cburgard:
[hs3.pdf](https://github.com/user-attachments/files/31913243/hs3.pdf)
After reading the section again less tired, I found it confusing and decided to refactor it once more.
Please let me know if this is okay to merge for you!

@oschulz @szabo137

oschulz:
I had ChatGPT Sol review this PR - the result is a bit lengthy, but very helpful, I think:

ChatGPT:

I would **request changes and not merge PR #101 as it stands**. The split between a Cartesian `product_dist` and an overlapping `point_product_dist` is a good direction, but several parts of the proposed semantics are currently inconsistent or underspecified in ways that could produce genuinely different statistical models across implementations. ([github.com][1])

The biggest problem is the new **analysis-context-dependent notion of what gets normalized**. That is much more consequential than it first appears.

## Findings

1. **Blocker: the proposed observable/parameter rule breaks priors.**

   The Analyses chapter explicitly allows the prior to be a `product_dist`. At the same time, the PR says that, in an analysis context, only variables interpreted as *observables* are integrated over, while parameters are held fixed. And the new analysis-context definition says variables constrained by supplied data are observables; the remaining variables are parameters.  

   Consider the completely ordinary Bayesian prior

   $$\pi(\theta_1,\theta_2) = \pi_1(\theta_1)\pi_2(\theta_2).$$

   Both $\theta_1,\theta_2$ are parameters of the analysis, so under the proposed rule the normalization set is empty. The proposed definition effectively gives

   $$M(\theta) = \int_{\varnothing} \pi_1(\theta_1)\pi_2(\theta_2)\,d\mu_\varnothing.$$

   With the usual $0$-dimensional product-measure convention,

   $$M(\theta)=\pi_1(\theta_1)\pi_2(\theta_2),$$

   and therefore

   $$ProductPdf(\theta) =\frac{\pi(\theta)}{M(\theta)} =1.$$

   So the prior disappears.

   Even if an implementation chose some other behavior for integration over an empty set, that's almost worse for a serialization standard: implementations would disagree.

   More fundamentally, Bayesian inference requires the same quantity $\theta$ to be a **parameter in**

   $$p(x\mid\theta)$$

   and a **random variate in**

   $$\pi(\theta).$$

   Its role cannot be assigned globally by variable name.

   Auxiliary constraints have the same issue. HS³ already permits distributions whose observed data are implicit rather than represented in the ordinary dataset list. 

   **Required fix:** normalization roles must at least be defined **per occurrence/use of a distribution**, rather than globally per variable name. Better still, make the mathematical denotation of the distribution itself context-independent.

2. **Blocker: this contradicts HS³'s existing foundational semantics.**

   The introduction currently presents HS³ using the standard forward-model picture

   $$m:\Theta\to\mathcal P(X),$$

   where fixing $\theta$ gives a concrete probability distribution over the distribution's variates. It also treats variates as part of the distribution's record structure. ([hep-statistics-serialization-standard.github.io][2])

   PR #101 effectively changes that to something closer to

   $$(m,\text{analysis context}) \longmapsto \text{probability distribution},$$

   because the context determines which named arguments are integrated out and hence what the normalized object actually is.

   That is not a small implementation detail. It changes the denotation of an HS³ object.

   There are two coherent designs:

   **A. Context-independent mathematical objects.** A distribution says intrinsically what its variates are. Conditional objects are explicitly probability kernels. This is what I recommend.

   **B. RooFit-style contextual normalization.** Then the standard needs to embrace that explicitly and revise the foundational model, priors, constraints, generation semantics, and evaluation semantics consistently.

   Putting the contextual rule only in the product/analysis sections leaves the specification internally inconsistent.

3. **Blocker: `point_product_dist` without `conditionals` is not generally identical to `product_dist`.**

   The PR says `product_dist` requires factor variate sets to be pairwise disjoint. It then introduces `point_product_dist` precisely to permit overlapping variates, but says:

   > Without `conditionals`, ... [it] is identical to the simpler Product distribution.

   Those statements cannot both hold. 

   For example,

   $$f(x)g(x)$$

   is a valid candidate for a globally normalized `point_product_dist`, but cannot be a `product_dist` under the newly proposed disjointness requirement.

   The correct statement is approximately:

   > Without `conditionals`, `point_product_dist` denotes the globally normalized pointwise product of its factors. If the factors' variate sets are pairwise disjoint, this reduces to `product_dist`.

   In fact, the no-conditional form is important for migrating old files that used overlapping factors with the old `product_dist`.

4. **Blocker: the worked conditional example changes the distribution.**

   This is mathematically important.

   Suppose the original product is

   $$p(x,y,z)\propto p_1(x,y)p_2(y,z)p_3(z).$$

   The PR conditionally normalizes $p_1$ over $x$, producing

   $$q_1(x\mid y) = \frac{p_1(x,y)} {Z_1(y)}, \qquad Z_1(y)=\int p_1(x,y)\,dx.$$

   The resulting raw product is therefore

   $$q_1(x\mid y)p_2(y,z)p_3(z) = \frac{ p_1(x,y)p_2(y,z)p_3(z) }{Z_1(y)}.$$

   Unless $Z_1(y)$ is constant in $y$, this is **not the original joint distribution**, even after a global normalization.

   RooFit's own documentation makes exactly this distinction: $F(x\mid y)G(y)$ and the globally normalized product $F(x,y)G(y)$ are in general different joint distributions. ([root.cern.ch][3])

   So the standard should not say the original product “may be represented” this way. It should say that the user is intentionally constructing the model

   $$p(x,y,z) = p_1(x\mid y)\,p_{23}(y,z),$$

   or equivalent.

5. **Blocker: reference-measure compatibility for overlapping variates is missing.**

   HS³ already defines densities in measure-theoretic terms: a PDF is a Radon–Nikodym derivative relative to an implied reference measure, such as Lebesgue measure for continuous distributions or counting measure for discrete ones. ([hep-statistics-serialization-standard.github.io][2])

   Once factors share a variate, multiplying densities only has a straightforward meaning if the factors use compatible reference measures on that coordinate.

   For example, naively writing

   $$f_{\rm Gaussian}(x)\, f_{\rm Poisson}(x)$$

   combines a density with respect to Lebesgue measure and a mass function with respect to counting measure. The numerical multiplication is possible; its interpretation as a density is not canonical.

   `point_product_dist` therefore needs a normative requirement along the lines of:

   > For every shared variate, all factors contributing density dependence on that variate MUST use mutually compatible reference measures.

   Or the composite must explicitly specify its reference measure and the necessary Radon–Nikodym conversions. The former is substantially simpler.

6. **Blocker: the conditional normalization needs a fully executable mathematical definition.**

   Right now prose is doing too much work. I would give the construct a single normative equation.

   Let $I$ be all factors, $\mathcal G$ the disjoint collection of conditional factor groups, and $R$ the remaining factors. For each group $G$,

   $$U_G = O\cap \left( \bigcup_{i\in G}V_i \right) \setminus C_G ,$$

   where $V_i$ is the factor's variate set, $C_G$ the declared conditioning set, and $O$ the normalization-variable set for this occurrence.

   Then

   $$Z_G = \int \prod_{i\in G}f_i\, d\mu_{U_G}, \qquad q_G = \frac{\prod_{i\in G}f_i}{Z_G},$$

   followed by

   $$h = \left(\prod_{G\in\mathcal G}q_G\right) \left(\prod_{i\in R}f_i\right), \qquad p = \frac{h} {\int h\,d\mu_O}.$$

   The spec then needs explicit rules for $Z_G=0$, $Z_G=\infty$, undefined integrals, empty $U_G$, null sets, and incompatible measures.

   In particular, **empty $U_G$** cannot be left to convention. RooFit has explicit behavior for terms that cease to depend on any normalized observable; HS³ should decide whether to copy that behavior or reject the construction. ([root.cern.ch][3])

7. **Important semantic distinction: normalized conditional factor ≠ conditional of the final joint.**

   Suppose

   $$q(x\mid y)$$

   is locally normalized in $x$, but another factor $r(x,y)$ also depends on $x$. Then after forming

   $$p(x,y)\propto q(x\mid y)r(x,y),$$

   generally

   $$p(x\mid y)\neq q(x\mid y).$$

   So the current construction can perfectly well define a **product of locally normalized factors followed by a global normalization**, but that is not automatically a chain-rule/Bayesian-network factorization.

   The standard should choose which concept it means.

   If chain-rule semantics are actually intended, I'd model the conditional pieces as **probability kernels**. Then a chain such as

   $$p(x,y,z) = p(x\mid y)\, p(y\mid z)\, p(z)$$

   has a clear generative meaning and is automatically normalized by iterated integration.

8. **The output variate set must be intrinsic or precisely inferred.**

   In something denoted

   $$q(x\mid y),$$

   what is $y$ as far as the serialized distribution is concerned?

   A conditioning/input coordinate? A variate? A model parameter? Something determined later by the analysis?

   The PR currently makes that depend too heavily on context. That's exactly the kind of thing two backends will eventually interpret differently.

   I'd distinguish explicitly between a probability distribution and a probability kernel:

   $$K:Y\to\mathcal P(X).$$

   Then $X$ are output/random coordinates and $Y$ are conditioning/input coordinates. Composition becomes mathematically clean.

9. **Extended-distribution behavior needs to be specified or forbidden.**

   `product_dist` and the new point product reference arbitrary distributions. What happens if one factor is extended? What if two are?

   Does the composite inherit an expected event count? Multiply counts? Ignore them? Reject the object?

   RooFit gives `RooProdPdf` defined behavior for extendable components; HS³ needs its own normative rule rather than letting backends inherit framework-specific behavior. ([ROOT][4])

   If this is out of scope for PR #101, the safe standard is:

   > Factors of `point_product_dist` MUST NOT be extended distributions.

   Then add support deliberately later.

10. **The schema-level details need another tightening pass.**

The nested object inside `conditionals` needs proper field definitions and cardinality/uniqueness rules, not merely an example. Define duplicate factor references, duplicate `on` entries, whether `on: []` is legal, unknown names, and behavior for implementations that don't support this feature.

Also, `product_dist.x` is described as a singular name even though its Cartesian product can plainly be multivariate, while `point_product_dist.x` is an array. That needs harmonization. And the JSON example currently omits the otherwise-required distribution `name`. 

11. **I'd change some terminology too.**

For `product_dist`, “product of independent distributions” is slightly backwards mathematically. Independence is a property of the resulting joint measure. More precise wording would be:

> the product measure / Cartesian product of the factor distributions.

This matters because factors can still share model parameters even though their **variates** are disjoint. They are conditionally independent given those parameters, not necessarily independent in every statistical interpretation.

12. **The compatibility story should be explicit.**

The current standard did not make the new pairwise-disjoint restriction on `product_dist` explicit. So PR #101 potentially turns previously accepted overlapping products into invalid documents.

That's defensible, but specify the migration:

$$\text{old overlapping `product_dist`} \quad\longrightarrow\quad \text{`point_product_dist` without conditionals}.$$

The PR title/description should also be brought into line with the current design; this is no longer merely “factorized normalization for `product_dist`.” ([github.com][1])

## What I would standardize

I'd keep `product_dist` extremely boring:

$$P = P_1\otimes\cdots\otimes P_n,$$

with pairwise-disjoint variate records. If each $P_i$ is already a probability distribution, no extra normalization constant is mathematically necessary:

$$\int \prod_i p_i(x_i)\, d(\mu_1\otimes\cdots\otimes\mu_n) = \prod_i\int p_i\,d\mu_i = 1.$$

That gives it a completely context-independent meaning.

Then make `point_product_dist` the **product-of-experts** construction:

$$p(x) = \frac{\prod_i f_i(x)} {\int\prod_i f_i(x)\,d\mu(x)},$$

allowing overlapping coordinates subject to compatible reference measures.

For conditional factorization, I would strongly prefer an explicit probability-kernel concept, whether it is a separate object or a very carefully defined part of `point_product_dist`. A kernel

$$K(dx\mid y)$$

has an intrinsic distinction between its normalized/output coordinates $x$ and conditioning coordinates $y$. That avoids inferring mathematics from whatever analysis happens to consume the object.

This also gives a principled representation for

$$p(x,y,z) = K_1(dx\mid y)\, K_2(dy\mid z)\, P_3(dz),$$

and gives implementations a natural sampling order. It connects the serialization directly to probability kernels, graphical models, and regular conditional probability rather than to one framework's normalization API.

If matching RooFit's dynamic normalization semantics is itself an explicit interoperability requirement, that's possible—but I would represent that as a distinct evaluation construct rather than redefine the meaning of “distribution” globally. RooFit's behavior is useful precedent, but the serialization standard should encode the mathematics, not serialize RooFit's evaluation state. ([root.cern.ch][3])

### Bottom line

The PR contains an important conceptual improvement—**separating Cartesian products from pointwise products is the right abstraction**. But I would block the present version on four things above all:

**the prior/parameter contradiction; the context-dependent denotation of distributions; the false `point_product == product_dist` equivalence; and the conditional example that silently changes the represented joint distribution.**

Those aren't wording nits. They affect which probability measure an HS³ document denotes.

For a standard, I think the key invariant should be:

*An HS^3 statistical object should have one mathematical denotation independent of the backend and, ideally, independent of its consumer.*

If that principle is adopted, most of the difficult choices in #101 become substantially easier.

[1]: https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/pull/101/files "Introduce factorized normalization for `product_dist` by cburgard · Pull Request #101 · hep-statistics-serialization-standard/hep-statistics-serialization-standard · GitHub"
[2]: https://hep-statistics-serialization-standard.github.io/chapters/1_introduction/ "Introduction - HEP Statistics Serialization Standard (HS³)"
[3]: https://root.cern.ch/doc/master/classRooProdPdf.html?utm_source=chatgpt.com "ROOT: RooProdPdf Class Reference"
[4]: https://root.cern.ch/doc/master/classRooProdPdf.html "ROOT: RooProdPdf Class Reference"

cburgard:
Thanks for the detailed review. I have revised the draft substantially. Below is a point-by-point explanation of the changes and the reasoning behind them.

1. **Priors and variable roles.** Removed the analysis-wide assignment of observables and parameters. Roles are now explicitly relative to a particular use: a quantity can be a parameter of a likelihood model and a variate of its prior. The prior is normalized over its variates, so its contribution does not disappear merely because those quantities are likelihood parameters.

2. **Context dependence and foundational semantics.** Added definitions of *variable*, *parameter*, and *variate/observable*, and clarified how roles are established for each use. We deliberately retain the ability to reuse a distribution with different choices of variates. The [response-model example](https://gist.github.com/cburgard/cd27b5037ad811fc9020eec5fefced20#file-response_model-json) illustrates the motivation: two likelihoods reference the same Gaussian, with `x: "reco"` and `mean: "truth"`, but pair it with data over `reco` and `truth`, respectively. The Gaussian is therefore normalized over different variables in the two uses. Consequently, we have not adopted intrinsically fixed variates or a separate kernel type. The requirement is that each use determine its variates and domain unambiguously; implementations must report underspecified uses rather than guess.

3. **Cartesian versus pointwise products.** Corrected the equivalence statement. A `point_product_dist` without conditional groups is a globally normalized pointwise product. It agrees with `product_dist` when the factors satisfy the Cartesian-product requirements. Overlapping factors remain a valid reason to use `point_product_dist` without `conditionals`.

4. **The conditional example.** Rewritten the example as an intentional construction of a new joint distribution. It now explicitly explains that conditionally normalizing the first factor divides the original product by a generally nonconstant $Z_1(y)$, changing the model.

5. **Reference measures.** Added a requirement that factor densities use the same reference measure for each shared variate in the particular use. Implicit conversion between incompatible measures is prohibited. This requirement concerns variates, not shared parameters.

6. **Complete normalization semantics.** Added equations defining the conditional groups, their integration variables, the remaining factors, and the final normalization. Parameters are excluded from integration regardless of whether they appear in `on`. When no variates remain in a group’s integral, its conditional factor is defined to be identically one, including where its factors vanish. For nonempty integration sets, normalizers must exist and be finite and strictly positive; the global normalizer must satisfy the same requirement. There is no implicit null-set exception. Undefined conditional densities require an appropriate domain restriction or an explicitly supplied conditional density.

7. **Local conditional factors versus conditionals of the final joint.** Added an explicit explanation that other factors can alter the final conditional density. The construction therefore does not generally imply a chain-rule factorization or an independent sampling procedure.

8. **Which quantities are variates.** Clarified that a variable named in `on` may be a variate in one use and a parameter in another. In the former case, it is excluded from the group’s integral but included in the complete product’s integral; in the latter, it is held fixed throughout. We use nondirectional terminology and retain explicit conditional language. Associated data or the statistical operation must establish the variates and domain.

9. **Extended factors.** Retained support because this is an important use case. Event-count terms are preserved separately from normalization of event-density shapes, and replacing a Cartesian product with a point product must preserve extended behavior wherever both constructions apply. Multiple extended factors may describe distinct, conditionally independent counts. Repeated contributions from the same count observation—including through nested factors—are prohibited. For different measurements of the same events, the example recommends combining the event densities and extending the composite once. Implementations must not silently remove, merge, or select competing count terms.

10. **Field definitions and implementation behavior.** Added explicit definitions for the nested `factors` and `on` fields, reference and uniqueness requirements, and empty-list behavior. An empty `conditionals` array means no conditional groups; an empty `on` array means conditioning on no variables. Removed `x` from both product types because it adds no independent meaning under the chosen semantics, and added the missing example `name`. Unsupported operations must be reported rather than implemented by silently dropping declarations or substituting another construction.

11. **Terminology and independence.** Defined `product_dist` as the Cartesian product of its factors. Factors may share parameters, but their variates must be disjoint, and no factor may depend on another factor’s variates, including through functions or fields such as `mean`. Independence is stated for fixed parameter values.

12. **Compatibility and release documentation.** Migration guidance will be recorded in the release changelog rather than the normative standard. This includes previously accepted overlapping products and historical `x` fields. The PR title and description also still need to be aligned with the final proposal.

The main deliberate difference from the recommended design is retaining variable roles that depend on the particular use. The revised text makes that choice explicit and constrains its interpretation, while preserving conditionality as the central language of the proposal.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,259 @@ If all coefficients are given, their sum is used as the expected number of event
 
 #### Product distribution {#sec:product-distribution label="sec:product-distribution"} 
 
-The product of independent distributions $f_i$ is defined as 
+A product distribution is the Cartesian product of its factor distributions.
```
szabo137, line 69: I know what you mean by "Cartesian product" here, but it isn't the correct definition of a product distribution. 

In the correct distribution, one speaks about the product of the random variables and "statistical independence" rather than the product of the PDFs of a set of random variables, e.g.

> A product distribution is a distribution constructed by the product of two or more statistically independent random variables. For two random variables X and Y, the PDF of Z = XY is given by $\int_0^\infty f_X(x)f_Y(z/x)/|x|dx$

Therefore, calling it a "product distribution" is highly misleading because there is a huge difference between the product of PDFs (what we have here) and the actual product distribution, which is the distribution resulting from a product of the random variables (where the resulting pdf is given by the Mellin transform above).

According to a short Google search, distributions formed by multiplying PDFs are usually called "normalized pointwise products" or "pointwise product distributions".

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,259 @@ If all coefficients are given, their sum is used as the expected number of event
...
+$$
 
+where each factor is normalized over its variates on the applicable domain,
+with the parameters held fixed.
+
+-   `name`: custom string
+-   `type`: `product_dist`
+-   `factors`: non-empty array of unique names referencing existing distributions
+
+The variates of the factors [must]{.smallcaps} be pairwise disjoint for
```
szabo137, line 85: What does "disjoint" mean in this context? Does this mean statistical independence of the corresponding random variables, or disjointness of the domains of the respective samples?

I assume the latter, because in this section, statistical independence is not used as a concept (nor are random variables).

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,259 @@ If all coefficients are given, their sum is used as the expected number of event
...
+this use. Each factor may depend on its own variates and on the parameters
```
szabo137, line 86: > Each factor may depend on its own variates 

Isn't that trivial?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -66,20 +66,259 @@ If all coefficients are given, their sum is used as the expected number of event
...
+described in Section [Distributions](#sec:distributions).
+
+For other types of products, please refer to the [Point-product distribution](#sec:point-product-distribution) or the [Product function](#sec:product-function).
+
+#### Point-product distribution {#sec:point-product-distribution label="sec:point-product-distribution"}
+
+A point-product distribution combines densities by pointwise multiplication,
+allowing the variates of its factors to overlap. Groups of factors may be
+used conditionally on specified variables before the complete product is
+normalized.
```
szabo137, line 106: According to my previous comment, this shall not be called "pointwise product" anymore. 

The whole question of these types of "products of distributions" boils down to these categories: 

1. A distribution of a (pointwise) product: $(X,f_X), (Y, f_Y) \rightarrow (X\times Y, (x,y) \mapsto f_X(x)*f_X(y))$, 
2. A (pointwise) product of distributions: $(X,f_X), (Y, f_Y) \rightarrow (Z, x\mapsto f_X(x)*f_Y(x))$, where both PDF must have the same domain. 
3. Distribution of the product of two random variables: $(X,f_X), (Y, f_Y) \rightarrow Z=XY, z\mapsto\text{Mellin}_{f_X, f_Y}(z)$, where $X$ and $Y$ must be statistically independent. 

I think the current "product distribution" is the first "distribution of a pointwise product," and the current "pointwise distribution" is the "pointwise distribution of products".

Unfortunately, I have no better names for these to propose yet.

docs/chapters/1_introduction.md
```diff
@@ -15,14 +17,30 @@ Please note that this document as well as the HS${}^3$ standard are still in dev
 
 ## Statistical semantics of HS${}^3$ {#sec:hs3-semantics} 
 
+A **variable** is a mathematical quantity whose value is chosen or generated,
```
szabo137, line 20: What does "mathematical quantity" mean in this context? Numbers? Or Vectors, or functions? Or anything that could represent a sample?

docs/chapters/1_introduction.md
```diff
@@ -15,14 +17,30 @@ Please note that this document as well as the HS${}^3$ standard are still in dev
...
+rather than directly computed from other quantities. A **parameter** is a
```
szabo137, line 21: What is the difference between "generated" and "computed"?

docs/chapters/1_introduction.md
```diff
@@ -15,14 +17,30 @@ Please note that this document as well as the HS${}^3$ standard are still in dev
...
+variable whose value is chosen by a user of the model, such as a minimizer.
+There is no expectation that a distribution is normalized with respect to
+its parameters, or that parameters themselves have specified statistical
+properties or a distribution from which they can be drawn. A **variate**,
+also called an **observable**, is a statistical quantity modeled by a
```
szabo137, line 26: What is a "statistical quantity"? What does "modelled by a distribution" mean in this context?

I think a cleaner introduction would be via "random variables" and their PDF, because they have a clear mathematical meaning (which must not be explained here), and they are easily mapped to your "variates" and "distributions". However, maybe this is a task for another PR. If you like, I can open an issue on that (or propose an introduction directly in a PR).

docs/chapters/1_introduction.md
```diff
@@ -15,14 +17,30 @@ Please note that this document as well as the HS${}^3$ standard are still in dev
...
+distribution; these two terms are used interchangeably in HS${}^3$.
+Both parameters and variates are variables. Their roles are relative to a
+particular use: the same variable may be a parameter in one use and a
+variate in another, for example a parameter of a likelihood model and a
+variate of a prior distribution.
```
szabo137, line 31: Does this mean they are not "well-defined" in the sense that a "thing" in HS3 could be a parameter or a variate, depending on the context in which it is used?

If so, it might not be a good idea to switch roles of objects depending on the context the object is used. Makes at least the desired implementation less predictable.

szabo137 (commented):
Sorry for the late reply, I added some comments on the clairification of words/definitions/concepts used. I am not sure if all of them must be resolved in this PR, but I would like to rather raise some awareness for them. 

However, I think for this PR it is critical to distinguish between different ways of multiplying distributions, as described in the comments.

docs/chapters/1_introduction.md
cburgard: I wouldn't use the word well-defined here, but yes, their role can depend on the context.
Please consider this example for reference: https://gist.github.com/cburgard/cd27b5037ad811fc9020eec5fefced20#file-response_model-json
I hope that helps to clarify the issue that this treatment is trying to solve.

docs/chapters/1_introduction.md
cburgard: Yes, that is true. I would propose the nomenclature
- random variable = observable = variate
- parameter
- variable = superset of (random variable, parameter)
Would you be fine with this?

docs/chapters/1_introduction.md
cburgard: The intent here was that "computed" is supposed to mean obtained deterministically as the result of an algorithm or function.
"Generated", on the other hand, was supposed to mean "drawn as an realization of a random variable".

docs/chapters/1_introduction.md
cburgard: This is primarily numbers (integer or real), but may include vectors in the future, as well as the results of function evaluations (not the function itself, that would not be a "quantity")
I'm not sure what "anything that could represent a sample" is supposed to encompass.

docs/chapters/2.1.2_composite_distributions.md
cburgard: That's unfortunate. I'm not married to any one naming scheme, I do think we need to come up with one though.

docs/chapters/2.1.2_composite_distributions.md
cburgard: Yes, but leaving it out also makes the sentence incomplete - otherwise, if we just write "Each factor may depend on the parameters", one would wonder if this word is meant to include the variates here, which it very much is not.

docs/chapters/2.1.2_composite_distributions.md
cburgard: Disjointness of the domains, but writing this explicitly seemed confusing, as $x\in[0,10]$ and $y\in[5,20]$ share a part of an interval, but are not defined in the same space/dimension. I was afraid that talking about domains here could be misinterpreted as implying that this is also forbidden.

docs/chapters/2.1.2_composite_distributions.md
cburgard: No objection, but for the naming, please refer to the second comment.

oschulz: > I know what you mean by "Cartesian product" here, but it isn't the correct definition of a product distribution.

Indeed, a Cartesian products is for sets, not (probability) measures.

docs/chapters/2.1.2_composite_distributions.md
oschulz: I agree - the names (in HS3 specifically) must be different, but the domains of the measures/distributions in the product can be anything - the equal, overlapping, or completely different. "disjoint" is probably not an ideal term here.

cburgard:
@oschulz @szabo137 @stalbrec I have decided to factorize the discussion on the roles of variables (variates, parameters) into a separate issue. Perhaps we can first resolve that point over there.

https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/issues/125

PR #103, opened 2026-07-07
HistFactory interpolation specification

stalbrec:
Refactors the interpolation section to use a structured interpolation component (`type`, `in`, `out`) with a default_interpolation fallback.

### DIFF
--- a/docs/parts/histfactory.md
+++ b/docs/parts/histfactory.md
@@ -77,7 +77,7 @@ The following functions are allowed for `in` and `out`:
 The choice of functions for `in` and `out`, together with the anchor conditions ($y(0) = y_0$, $y(\pm1) = y_\pm$) and, where required, matching of first and second derivatives at the boundary ($\left.\frac{\mathrm{d}y}{\mathrm{d}\theta}\right|_{\theta=\pm 1}$ and $\left.\frac{\mathrm{d}^2y}{\mathrm{d}\theta}\right|_{\theta=\pm 1}$), fixes all free parameters of the chosen functions except for $\theta$ itself.
 
 
-The following combinations are often used and showcase the :
+The following choices are often used and showcase the resulting functional forms for common combinations of `type`, `in` and `out`:
 
 - additive, piecewise linear: 
     - `{"type":"add", "in":"poly1", "out":null}`
--- a/docs/parts/histfactory.md
+++ b/docs/parts/histfactory.md
@@ -98,7 +98,8 @@ The following choices are often used and showcase the resulting functional forms
 - multiplicative, polynomial (6th degree) interpolation and exponential extrapolation 
     - `{type:"mult", in:"poly6", out:"exp"}`
     - with $y_5(\theta, y_0, y_+, y_-) = \left\{\begin{array}{ll} (y_+/y_0)^\theta &  \theta\geq 1 \\ 1+\theta\cdot(a+\theta\cdot(b+\theta\cdot(c+\theta\cdot(d+\theta\cdot(e+\theta\cdot f)))))& |\theta|<1\\ (y_-/y_0)^{-\theta} &  \theta\leq-1\end{array}\right.$,  
-  where: $\begin{array}{l}a=\tfrac18(15A_0-7S_1+A_2) \\ b=\tfrac18(-24+24S_0-9A_1+S_2) \\ c=\tfrac14(-5A_0+5S_1-A_2) \\ d=\tfrac14(12-12S_0+7A_1-S_2) \\ e = \tfrac18 (3A_0 -3S_1+A_2) \\ f=\tfrac18 (-8+8S_0-5A_1+S_2)\end{array}$, with: $\begin{array}{l}S_0=\tfrac12(y_+ + y_-) \\ A_0=\tfrac12(y_+ - y_-) \\ S_1=\tfrac12(y_+\log{y_+} + y_-\log{y_-}) \\ A_1=\tfrac12(y_+\log{y_+} - y_-\log{y_-}) \\ S_2=\tfrac12(y_+\log^2{y_+} + y_-\log^2{y_-}) \\ A_2=\tfrac12(y_+\log^2{y_+} - y_-\log^2{y_-})\end{array}$ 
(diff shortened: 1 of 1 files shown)


Phmonski:
While implementing the serialization mapping in RooFit, I noticed three apparent formula/documentation issues in this PR:

1. The additive `poly6` coefficient is written as `A = 1/6`, while RooFit and the established formula use `A = 1/16`.
2. The negative-side `poly2` extrapolation appears to use `2s+d`; continuity of the derivative requires `2s-d`.
3. The general response is described as being anchored at `y(0)=y0`, while the combination equations and examples require additive responses to be anchored at zero and multiplicative responses at one.

stalbrec:
Thank you very much for your feedback! :)
I fixed the typos for your points 1 & 2.
For point 3, I'm not sure that I understand what you mean. Are you saying for e.g. `"type":"mult"` we need to require $y_0=1$ ? 
If so, then I am missing to see why that is.

cburgard:
> Thank you very much for your feedback! :) I fixed the typos for your points 1 & 2. For point 3, I'm not sure that I understand what you mean. Are you saying for e.g. `"type":"mult"` we need to require y 0 = 1 ? If so, then I am missing to see why that is.

@Phmonski @stalbrec is this point resolved now, or is there still an open question attached to point 3?

Phmonski:
Interpolation schemes are used not only by HistFactory normsys and histosys modifiers, but also by the standalone RooFit classes FlexibleInterpVar and PiecewiseInterpolation. I have already implemented structured HS3 import and export for both in [ROOT PR #22844](https://github.com/root-project/root/pull/22844), specifically in [this commit](https://github.com/root-project/root/commit/58b6beb2481e284fef18132b9317f1edb6b89acb).

The implemented representations use:
- type: "interpolation0d" for FlexibleInterpVar;
- type: "interpolation" for PiecewiseInterpolation;
- an interpolations array containing the same structured {type, in, out} descriptors introduced here.

If interpolations contains exactly one descriptor, it applies to all parameters. Otherwise, it must contain exactly one descriptor per parameter, positionally aligned with vars, low, and high. ROOT exports only the structured representation, while still accepting the legacy integer interpolationCodes array on import.

Phmonski:
> Thank you very much for your feedback! :) I fixed the typos for your points 1 & 2. For point 3, I'm not sure that I understand what you mean. Are you saying for e.g. `"type":"mult"` we need to require y 0 = 1 ? If so, then I am missing to see why that is.

The inconsistency is between [line 56](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/blob/e36cc7c7ae19cbe020351c470830b5fe478c0b91/docs/parts/histfactory.md?plain=1#L56), which defines each response by $y(0)=y_0$, and the combination equations on [lines 58–68](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/blob/e36cc7c7ae19cbe020351c470830b5fe478c0b91/docs/parts/histfactory.md?plain=1#L58-L68), which add or multiply those responses with another $y_0$. The concrete definitions on [lines 82–105](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/blob/e36cc7c7ae19cbe020351c470830b5fe478c0b91/docs/parts/histfactory.md?plain=1#L82-L105) instead show the intended convention: additive responses are anchored at zero and multiplicative responses at one. Thus, I am not proposing $y_0=1$; I am proposing that the specification distinguish the absolute nominal prediction $Y_0$ from the normalized modifier response. 

Furthermore, there is also a related issue on [lines 100–105](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/blob/e36cc7c7ae19cbe020351c470830b5fe478c0b91/docs/parts/histfactory.md?plain=1#L100-L105): the multiplicative `poly6` interior coefficients and 1+y₄(...) should use normalized anchors $y_\pm/y_0$ if arbitrary $y_0$ is supported.

@stalbrec

stalbrec:
Alright; i think partially the/my confusion stems from me ignorantly reusing $y(..)$ for the modified prediction as well as modifying response function. 
I renamed $y(\theta; y_-, y_0, y_+)$ to $I(\theta; y_-, y_0, y_+)$ and fixed up the description, such that its clear that $I$ is anchored to the respective normalized/unnormalized values of $y_\pm$ and $y_0$.
Does this sound/look reasonable?

In the process another bug was uncovered in the definition of $I_5$ which i also fixed.

stalbrec:
> Interpolation schemes are used not only by HistFactory normsys and histosys modifiers, but also by the standalone RooFit classes FlexibleInterpVar and PiecewiseInterpolation. I have already implemented structured HS3 import and export for both in [ROOT PR #22844](https://github.com/root-project/root/pull/22844), specifically in [this commit](https://github.com/root-project/root/commit/58b6beb2481e284fef18132b9317f1edb6b89acb).
> 
> The implemented representations use:
> 
> * type: "interpolation0d" for FlexibleInterpVar;
> * type: "interpolation" for PiecewiseInterpolation;
> * an interpolations array containing the same structured {type, in, out} descriptors introduced here.
> 
> If interpolations contains exactly one descriptor, it applies to all parameters. Otherwise, it must contain exactly one descriptor per parameter, positionally aligned with vars, low, and high. ROOT exports only the structured representation, while still accepting the legacy integer interpolationCodes array on import.

I propose we add these in a later PR, which will also move the description of common Interpolationfunctions into an appendix both `histfactory_dist` and the `interpolation(0d)` functions can point to. What do you think @Phmonski ?

Phmonski:
> I propose we add these in a later PR, which will also move the description of common Interpolationfunctions into an appendix both `histfactory_dist` and the `interpolation(0d)` functions can point to. What do you think @Phmonski ?

This is already live on RooFit so I suggest we come up with a PR in the near future

oschulz:
Corresponding table in FlatPPL (correct?): https://flatppl.github.io/flatppl-design/#three-point-interpolation-functions

stalbrec:
Hi @oschulz
I had a look at the table and do think we are in agreement.

For reference: this PR adds the RooFit interpolation codes defined in [MathFuncs.h](https://github.com/root-project/root/blob/master/roofit/roofitcore/inc/RooFit/Detail/MathFuncs.h#L254-L366). 
The codes/naming maps as follows:


RooFit | pyhf | hs3 | flatppl
-- | -- | -- | --
0 | code0 | add/poly1/null | interp_pwlin
1 | code1 | mult/exp/null | interp_pwexp
2 | code2 | add/poly2/poly1 | interp_poly2_lin
3 n/a |   |   |  
4 | code4p | add/poly6/poly1 | interp_poly6_lin
5 | code4 | mult/poly6/exp | interp_poly6_exp
6 |   | mult/poly6/poly1 |

oschulz:
Thanks for checking @stalbrec !

PR #105, opened 2026-07-13
Add categorical axes, combined data sets, indexed distributions, and diagonal Gauss/Poisson constraint distributions

stalbrec:
Adds categorical (discrete-valued) axes as a new axis kind in unbinned/binned data and product domains, plus three constructs built on them:

- **`combined`** data set: organizes sub-datasets of any type along named categorical dimensions
- **`indexed_dist`**: sub-distributions selected by a categorical `index`. With `exclusive: false` the index is a data axis and all members contribute jointly (`RooSimultaneous`); with `exclusive: true` it is a categorical parameter and only the active member is evaluated, with an optional `correction` term matching `RooMultiPdf`'s discrete-profiling `cFactor`.
- **`diagonal_gauss_dist` / `diagonal_poisson_dist`**: flat products of independent 1D constraint terms indexed against a categorical data axis, declared as one construct so engines can vectorize. 

A section on auxiliary measurements and global observables within the likelihoods chapter recommends pairing a diagonal distribution with a categorical-axis data set as the compact encoding for many same-kind constraint terms - addressing https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/discussions/20


docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -61,7 +61,78 @@ If all coefficients are given, their sum is used as the expected number of event
...
 -   `coefficients`: array of names of coefficients $c_i$ or numbers
--   `extended`: boolean denoting whether this is an extended model
+-   `extended`: boolean denoting whether this is an extended model  
+
+
+#### Indexed distribution
+
+The indexed distribution combines severeal sub-distributions selected by the value of a discrete/categorical variabel `index`. `index` [may]{.smallcaps} refer either to a categorical axis of the data this distribution is paired with, or to a parameter with a categorical domain. Which of the two determines how the distribution is evaluated.
+
+If `index` refers to a data axis (`exclusive: false`), evaluation is dispatched per entry:
```
cburgard, line 71: I don't understand what the member `exclusive` is supposed to convey.
From what I read here, its only role is to distinguish whether `index` is an observable (data axis) or a parameter. This distinction is something that we generally try to avoid. Why is it needed here? Does it actively change the mathematics in a meaningful way?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -61,7 +61,78 @@ If all coefficients are given, their sum is used as the expected number of event
...
+    - `categories`: array of unique strings enumerating the category values used by this distribution
+- `members`: array of structs, each assigning one category value to the distribution used for it:
+    - `category`: string, [must]{.smallcaps} match to one of `index.categories`
+    - `dist`: name of the sub-distribution used for this value of `category`
+- `exclusive`: ([optional]{.smallcaps}) boolean, defaults to `false`. If `true`, `name` refers to a parameter, and only the single member matching that parameter's current value is evaluated. If `false` (default), `name` refers to a categorical axis of the paired data, and every member contributes jointly, dispatched per data entry.
+- `correction`: ([optional]{.smallcaps}) number used as additional correction to the $-2\log{\mathcal{L}}$ contribution when this member is the active choice. Defaults to $0.5*n_p$ (`correction: 0.5`) with the number of parameter $n_p$ of the active member distribution (free model parameters + observables). Only meaningful when `exclusive` is `true`, can be used to penalize the different model complexity between the candidate member distributions.
```
cburgard, line 103: I am confused by this `correction` member. The way I understand the text, it refers to a multiplication with a fixed value "when this member is the active choice".
That seems to imply that this is a property of one of the members. However, in your example further down, the `correction` is a member of the distribution itself. Which one is it? Please make sure that is uniquely determined by the description.

Also, on a more fundamental level: What exactly is the mathematical role of this correction?
The members of the distribution are themselves distributions. I assume that multiplication refers to a multiplication with the density value here rather than a scaling factor for one of the variates or parameters. That seems confusing to me, since distributions are normalized - hence any multiplicative factor is going to be removed by the integration constant.
Also, even if this is what it's meant to do: how is it different from wrapping the distribution in a `product` with a fixed factor of 0.5? The latter seems a cleaner and less convoluted option...

Trying to read between the lines of your proposal now, it seems that this is meant to affect the NLL value if and only if this is the top-level distribution of a graph, in order to model the factor two of the "- 2 log (L)". That, however, should be a under the discretion of the engine to decide how to define the function it is minimizing - at least in the way I see it. Perhaps there is a good reason for this not to be the case, but I don't see it at the moment.

That leaves me to wonder: What exactly is the role and importance of that factor? Does it really add a meaningful distinction? If so, which one?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -61,7 +61,78 @@ If all coefficients are given, their sum is used as the expected number of event
...
+
+The distributions referenced by `members` [may]{.smallcaps} each have different variate names, since only one member's distribution is evaluated for any given category value. `members` [need not]{.smallcaps} cover every entry of `index.categories`; categories without a matching entry are undefined for this distribution. The 
+
+```json title="Example: Indexed distribution, joint"
+{
+    "name":"combined_pdf",
+    "type":"indexed_dist",
+    "index":{"name":"channel", "categories":["SR","CR"]},
+    "members":[
+        {"category":"SR", "dist":"sr_pdf"},
```
cburgard, line 113: In our examples, I think we generally do not use the `pdf` suffix.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +193,101 @@ The distribution can be specified either via the non-normalized density function
...
+where $d_i$ is read from the column of the paired data set whose `index` category matches the corresponding entry of `member`. It is sematically equivalent to a [Product distribution](#sec:produc-distribution){referecen-type="ref" reference="sec:product-distribution"} of $n$ one-dimensional Gaussians, but declared flat and explicit such that engines [may]{.smallcaps} use a dedicated, vectorized implementation.  
+
+A DiagonalGauss distribution is a struct defined by the following components:  
+
+- `name`: custom string
+- `type`: `diagonal_gauss_dist`
+- `index`: struct describing the categorical variable that labels each auxiliary measurement of $d$
+    - `name`: name of the categorical variable. [must]{.smallcaps} refer to a categorical axis of the data this distribution is paired with
+    - `categories`: array of unique strings enmerating the category values used by this distribution
+- `d_axis`: name of the continuous axis of the paired data holding $d_i$, read from the column matching the category of `index`
```
cburgard, line 218: I don't understand this member `d_axis`. From what I understand, this is supposed to name the continuous axis of the gaussian distributions that are lined up here along the index category.
I don't think we generally use the prefix `d` to signify data. Also, again, we generally try to not identify the roles of observables and parameters within the context of a single distribution. What is more, it would be very good if we could try and be consistent in naming with the existing gaussian distribution(s) (see above).
Also, none of your examples use this member. Why not?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +193,101 @@ The distribution can be specified either via the non-normalized density function
...
+\text{DiagonalGaussPdf}(\vec{d}) &= \prod_{i=1}^{n} \text{GaussianPdf}(d_i; \mu_i, \sigma)
+\end{aligned}
+$$
+
+where $d_i$ is read from the column of the paired data set whose `index` category matches the corresponding entry of `member`. It is sematically equivalent to a [Product distribution](#sec:produc-distribution){referecen-type="ref" reference="sec:product-distribution"} of $n$ one-dimensional Gaussians, but declared flat and explicit such that engines [may]{.smallcaps} use a dedicated, vectorized implementation.  
+
+A DiagonalGauss distribution is a struct defined by the following components:  
+
+- `name`: custom string
+- `type`: `diagonal_gauss_dist`
```
cburgard, line 214: For most other variants of a gaussian, we also offer a version of the type that is called `normal`, so I suggest to also call this `diagonal_normal_dist` as a second variant of the type (see e.g. `normal_dist` and `multivariate_normal_dist`).

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +193,101 @@ The distribution can be specified either via the non-normalized density function
...
+\end{aligned}
+$$
+where $d_i$, the fixed observed auxiliary count, is read from the row of the paired data set whose `index` category matches the corresponding entry of `members`.  
+
+- `name`: custom string
+- `type`: `diagonal_poisson_dist`
+- `index`: struct describing the categorical variable that labels each entry:
+    - `name`: name of the categorical variable. [must]{.smallcaps} refer to a categorical axis of the data this distribution is paired with
+    - `categories`: array of unique strings enumerating the category values used by this distribution
+- `d_axis`: name of the continuous axis of the paired data holding $d_i$, read from the column matching the category of `index`
```
cburgard, line 271: See above. Again, the naming is unconventional.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +193,101 @@ The distribution can be specified either via the non-normalized density function
...
+{
+    "distributions":[
+        ...
+            {
+                "name":"gauss_constraints",
+                "type":"diagonal_gauss_dist",
+                "index":{"name":"np", "categories":["np_1_in","np_2_in"]},
+                "mean_axis":"nom_np",
+                "members":[
+                    {"category":"np_1_in","x":"np_1"},
```
cburgard, line 234: It's not clear to my why, in this example, you label the category as `np_1_in`. What is the suffix `_in` supposed to indicate? "Index"? In this case, I think `_cat` (for "category") would be more appropriate.

Also, in the definition text above, you introduce the members as having `category` and `mu`, whereas here, you use them as `category` and `x`. This is inconsistent. I think the former (`mu` rather than `x`) makes more sense.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +193,101 @@ The distribution can be specified either via the non-normalized density function
...
+- `name`: custom string
+- `type`: `diagonal_gauss_dist`
+- `index`: struct describing the categorical variable that labels each auxiliary measurement of $d$
+    - `name`: name of the categorical variable. [must]{.smallcaps} refer to a categorical axis of the data this distribution is paired with
+    - `categories`: array of unique strings enmerating the category values used by this distribution
+- `d_axis`: name of the continuous axis of the paired data holding $d_i$, read from the column matching the category of `index`
+- `sigma`: ([optional]{.smallcaps}) number giving $\sigma$, shared by every member. Defaults to `1.0`
+- `members`: array of structs, each assigning one categorical value to the parameter it maps to $mu_i$
+    - `category`: string, [must]{.smallcaps} match one of `index.categories`
+    - `mu`: name of the mean $\mu_i$, can a parameter name or function.
```
cburgard, line 222: This distribution type is, as far as I understand, intended to give an efficient way to represent arrays of constraint terms.
This means that the quantity constrained here is generally understood to be the parameter to be constrained. Perhaps it would be worthwhile to hint at this in the text to make the usage less obscure.
Also, I think in our usual normal distribution, we use `mean` rather than `mu`. I think it would be good to be consistent there.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +193,101 @@ The distribution can be specified either via the non-normalized density function
...
+
+- `name`: custom string
+- `type`: `diagonal_poisson_dist`
+- `index`: struct describing the categorical variable that labels each entry:
+    - `name`: name of the categorical variable. [must]{.smallcaps} refer to a categorical axis of the data this distribution is paired with
+    - `categories`: array of unique strings enumerating the category values used by this distribution
+- `d_axis`: name of the continuous axis of the paired data holding $d_i$, read from the column matching the category of `index`
+- `members`: array of structs, each assigning one category value to the parameter it maps to $\lambda_i$:
+    - `category`: string, [must]{.smallcaps} match on of `index.categories`
+    - `lambda`: name of the rate $\lambda_i$, can a parameter name or function.
```
cburgard, line 274: Checking with our regular Poisson distribution, it seems we are using `mean` rather than `lambda` there. I think it would be good to be consistent and also switch to `mean` here.

docs/chapters/2.3_data.md
```diff
@@ -64,30 +67,48 @@ Unbinned data describes a measurement of multiple data points in a possibly mult
...
 ``` 
 
+```json title="Example: Unbinned Data with categorical axis"
+"data":[ 
+  { 
+    "name":"data2", 
+    "type":"unbinned", 
+    "weights":[ 12.0, 4.3, 2.1 ], 
+    "entries":[ [0,7,0.1], [0,9,-0.2], [1,4,0.9]], 
+    "entries_uncertainties":[ [0.1], [0.3], [0.4]],
```
cburgard, line 77: Let's try to not overcomplicate things here. Maybe get rid of the "uncertainties" in this example, as this is a somewhat complicated case that is [being overhauled at the moment](https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/discussions/88) anyway.

docs/chapters/2.3_data.md
```diff
@@ -64,30 +67,48 @@ Unbinned data describes a measurement of multiple data points in a possibly mult
 ]
 ``` 
 
+```json title="Example: Unbinned Data with categorical axis"
+"data":[ 
+  { 
+    "name":"data2", 
+    "type":"unbinned", 
+    "weights":[ 12.0, 4.3, 2.1 ], 
+    "entries":[ [0,7,0.1], [0,9,-0.2], [1,4,0.9]],
```
cburgard, line 76: I'm not sure I fully understand what you propose above how the categorical values should be represented in `entries`, but it seems you are not actually doing that in this example. Perhaps you can adjust this to include the categorical indices here?

docs/chapters/2.3_data.md
```diff
@@ -120,3 +158,44 @@ This type can also be used to store pre-processed data utilizing the `uncertaint
   ... 
 ] 
 ``` 
+
+### Combined data set
```
cburgard, line 162: This is perhaps being a bit nitpicky, but the distribution you introduce that maps most cleanly to this type of dataset is the "Indexed distribution". Why did you chose to then call the dataset "combined" rather than "indexed", or alternatively, the distribution "indexed" rather than "combined"?
I think it would be beneficial to have a consistent nomenclature, but I don't have a strong preference which one I prefer.

docs/chapters/2.3_data.md
```diff
@@ -120,3 +158,44 @@ This type can also be used to store pre-processed data utilizing the `uncertaint
...
+
+Combined data sets merge multiple data structs into one. The sub-data sets can be of any `type` described above and [may]{.smallcaps} have overlapping or non-overlapping `axes`.  
+A combined data set is defined with the components:  
+
+- `name`: custom string
+- `type`: `combined`
+- `indices`: array of structs defining the named dimensions spanning the collection of sub-data sets. Each struct [must]{.smallcaps} have the components `name` and `categories`, the latter being an array of the allowed values along that dimension, in the same way as a categorical axis.
```
cburgard, line 169: What you call `indices` here is called `dataset_axes` in your example. Neither seem particularly meaningful to me. Why not just use `axes`?

docs/chapters/2.4_likelihoods.md
```diff
@@ -43,3 +43,48 @@ The components of a likelihood struct are:
   ... 
 ] 
 ``` 
+### Auxiliary measurements and global observables
```
cburgard, line 46: I really like this section, I think it's very useful. However, I'm thinking whether this (rather explanatory) section would potentially profit from a "non-normative" label?

cburgard (commented):
Hi @stalbrec , thank you very much for this long-needed extension of the standard.
This accomplishes many things at the same time, namely the introduction of simultaenous pdfs, combined datasets and index category variables. Thank you very much for putting in so much work and effort into this!

As this is a crucial extension and improvement, I think it deserves a fair bit of thought and scrutiny, and while you clearly already put in a lot of effort, I did notice a few inconsistencies and/or have comments and questions on individual attributes and/or their naming.

Please find them as comments below. I would also like to invite other stakeholders (specifically @oschulz @kratsg ) to comment!

stalbrec:
I'm splitting this PR into three separate PRs, as requested.
If you want to give feedback or comment on something please do it in the respective new PR:

- Adding categorical axes to data and domain sections: #108 
- Adding dedicated types for combined data sets and combined distributions: #119
- Adding multidimensional (diagonal) Gauss/Poisson distributions to group aux measurements for constraints in categorical dataset: #109

docs/chapters/2.3_data.md
stalbrec: in this example the first column in the entries is the index of the respective category in the `categories` array:

For example the first entry `[0, 7, 0.1]` -> `0 -> SR`, `variable1=7`, `variable2=0.1`

docs/chapters/2.3_data.md
stalbrec: removed that part in #108

docs/chapters/2.1.2_composite_distributions.md
stalbrec: adressed in #109

docs/chapters/2.1.2_composite_distributions.md
stalbrec: adressed in #109

docs/chapters/2.1.2_composite_distributions.md
stalbrec: adressed in #109

docs/chapters/2.1.2_composite_distributions.md
stalbrec: Sorry for these confusing mismatches, you are right. These are mistakes I made during late careless changes.
adressed in #109

docs/chapters/2.1.2_composite_distributions.md
stalbrec: adressed in #109

docs/chapters/2.1.2_composite_distributions.md
stalbrec: adressed in #109

docs/chapters/2.4_likelihoods.md
stalbrec: added in #109

cburgard:
This PR has been superseded by 3 PRs mentioned by @stalbrec above, closing.

PR #106, opened 2026-07-14
Updated the README typography

cburgard:
The README was using slightly inconsistent heading levels and URL syntax. This PR fixes this.


BJMCox:
For me on GitHub this does not render as intended
image

Perhaps we should use the unicode ³ ?

PR #108, opened 2026-07-22
Add categorical axes to data and domains

stalbrec:
> This is PR 1/3 in an effort to split up #105 into smaller, independently reviewable pieces. This part covers categorical axes for `domains` and `data`; the remaining two will follow as separate PRs.

In addition to continuous observables, many analyses distinguish between discrete states of an experiment: selection regions, tagging categories etc.  
Furthermore, there are use cases of discrete parameters, e.g. in "discrete profiling" with an index parameter which is being iterated by the minimizer itself to choose the best fitting distribution from a familiy of distributions (as done with `RooMultiPDF` in RooFit).  

So far HS³ had no dedicated way to express such discrete-valued variables, as every axis is implicitly continuous, so categorical structure has to be reproduced e.g. by declaring one distribution/data pair per category, multiplying boilerplate for every additional channel.  

Categorical axes close this gap as a first-class counterpart to continuous axes. Both share the same structural role wherever an `axes` component appears. For now this is added for data sets and domains, such that a single dataset can span several named categories explicitly with a shared subset of axes. 

Categorical axes will further allow us 
1. to add types to HS³ that combine both data sets and distributions in a more generic way, in the sense that we can allow for combined  data sets and distributions with (partially) disjunct sets of continuous axes. This can cover the two related cases from above: "discrete profiling" and multi-channel combinations.   
Conceptionally:  $\text{PDF}_\text{combined}(x, c) = \text{PDF}(m_c(\theta), x)$ with $c$ either a free parameter (one of the sub-distributions is evaluated) or an observable (evaluated per data entry, jointly over $x$ and $c$) (see PR 2/3; TBD)
3. and to writing constraints/auxiliary measurements in a more compact way to HS$^3$ files than the current way (e.g. the way we save RooFits 'global observables').
These two will be addressed in upcoming PRs. (see PR 3/3; TBD)

#### What Changed
- `2.5_domains.md`: an axis in a `product_domain` can now be continuous (with `min`/`max` components) or categorical (with the `categories` as an array of unique strings)
- `2.3_data.md`: In the *Axes Specfication* subsection the existing axis is split into a *Continuous Axes* and a *Categorical Axes* subsections and the per data type rules are updated to state where categorical axes are allowed and how they are encoded:
	- `unbinned`: the corresponding column of `entries` holds the zero-based index of the row's category
	- `binned`: the axis acts as its own binning, contributing one bin per category to the flattened `contents` array.
- `1_introduction.md`: clarify that categorical axes are the exception to the 'parameter tuples must be flat and only consist of real numbers' rule.


docs/chapters/1_introduction.md
```diff
@@ -17,7 +17,7 @@ Please note that this document as well as the HS${}^3$ standard are still in dev
...
+Parameters in HS${}^3$ are always named, so semantically the elements of a parameter space $\Theta$ are records $\theta$ (i.e. tuples in which every entry has a name). Even if there is only a single free parameter, $\theta$ should be thought of as a single-entry record. In the current version of this standard, parameter tuples must be flat and only consist of real numbers, vector-valued or nested entries are not supported yet. The exception for this rule are categorical axes, which support discrete-valued variables, such as an analysis channel or a tagging state (see [Axis Specifications](#sec:axes){reference-type="ref" reference="sec:axes"}). Future versions of the standard will likely be less restrictive, especially with respect to discrete or vector-valued parameters. Parameter domains defined as part of this standard may be of various types, even though the current version only supports product domains. Future versions are likely to be less restrictive in this regard.
```
BJMCox: I think that the additional sentence in the introduction about the categorical axes is belabouring a point that has not yet been made. 

Also: ...of real numbers, vector-valued or ... -> ...of real numbers; vector-valued or... (or :, I forget which, but not ,)

docs/chapters/2.3_data.md
```diff
@@ -19,39 +19,52 @@ While data, in most settings, has no uncertainty attached to it, this standard *
...
 -   `min`: lower bound of the observable domain
 -   `max`: upper bound of the observable domain
 
 These fields define the domain over which PDFs are normalized.
 
-**Note on normalization**: Any PDF depending on observable $x$ must be normalized over the domain $[\text{min}, \text{max}]$ defined by that observable's axis, not over $(-\infty, \infty)$.
+**Note on normalization**: Any PDF depending on a continuous observable $x$ must be normalized over the domain $[\text{min}, \text{max}]$ defined by that observable's axis, not over $(-\infty, \infty)$.
+
+#### Categorical Axes
+A categorical axis describes a discrete-valued observable that takes one of a finite set of named states. A categorical axis [must]{.smallcaps} include:
```
BJMCox, line 36: Is this basically an enum? In R we would call this a 'factor' (although). It might be worth noting the relation if possible (maybe not to R though 🥲). 

How does this translate behind the scenes to ROOT/RooFit - in R factors are basically a mapping stored alongside an integer vector, I presume it would be similar here?

docs/chapters/2.3_data.md
```diff
@@ -19,39 +19,52 @@ While data, in most settings, has no uncertainty attached to it, this standard *
...
 -   Binning fields (`nbins`, `edges`) [must not]{.smallcaps} be present
+-   Categorical axes are permitted; the corresponding column of `entries` holds zero-based indices of the respective row's category within the `categories` array
+
+**For Binned Data:**  
 
-**For Binned Data:**
--   Base fields (`name`, `min`, `max`) are required
--   Exactly one binning specification [must]{.smallcaps} be present:
+-   Continuous axes are required to carry base fields (`name`, `min`, `max`)
+-   Each continuous axis [must]{.smallcaps} carry exactly one binning specification:
```
BJMCox, line 62: I would emphasise exactly one

docs/chapters/2.3_data.md
```diff
@@ -19,39 +19,52 @@ While data, in most settings, has no uncertainty attached to it, this standard *
...
     1. Regular binning: `nbins` (integer number of equal-width bins)
     2. Irregular binning: `edges` (array of length $n+1$ bin boundaries)
 -   Both `nbins` and `edges` [must not]{.smallcaps} be specified simultaneously
--   Neither `nbins` nor `edges` [must not]{.smallcaps} be omitted
+-   Neither `nbins` nor `edges` [must not]{.smallcaps} be omitted on a continuous axis
```
BJMCox, line 66: I do not understand what this means as written. I think what is meant is that only one of `nbins` and `edges` can be given, and that giving both of `nbins` and `edges` is not permitted.

If this is the case I would propose:
Only one of `nbins` and `edges` can be specified

docs/chapters/2.3_data.md
```diff
@@ -124,13 +154,13 @@ Binned data describes a histogram of data points with bin contents in a possibly
...
+    "name":"data3",
     "type":"binned",
     "contents":[ 9.0, 18.4 ],
     "axes":[ { "name":"variable1", "nbins":2, "min":1, "max":3 } ]
   },
   {
-    "name":"asimov_data2",
+    "name":"asimov_data3",
     "type":"binned",
     "contents":[ 9.0, 18.4, 13, 0. ],
```
BJMCox, line 165: I am sure this is valid (given we accept only reals), but it does look odd. If it is intentional I would be minded to comment on it

docs/chapters/2.3_data.md
cburgard: It is basically an enum, but one that should be treatable as an integer-valued observable ("variate") and/or parameter.

I don't know about R, but in ROOT, this maps fairly directly to this class/construct: https://root.cern/doc/master/classRooCategory.html

BJMCox: Excellent!

docs/chapters/2.3_data.md
stalbrec: I stumbled across this as well, but I did not want to mix this change into this PR.
I'll note this down and address this in an upcoming PR, if this is okay with you.

docs/chapters/2.3_data.md
stalbrec: You mean to emphasize using formatting (e.g. italics)?

docs/chapters/2.3_data.md
stalbrec: I agree this looks odd. I would address this in an upcoming PR (see above) to not mix it into this, since not necessarily related to this PRs topic, okay?

docs/chapters/2.3_data.md
BJMCox: That would work I think (on reflection it is probably fine as is 😊)

docs/chapters/1_introduction.md
stalbrec: Thank you very much for your feedback across the PRs! :)

I am not sure I agree (or understand) completely. Though, I am not super happy with this sentence and *now* think it reads awkward, I do think its necessary to mention this point or at least soften the rule on how parameter tuples should appear in the standard.

Maybe changing it to something like:
> In the current version of this standard, parameter tuples must be flat and only consist of real numbers (or strings for categorical parameters); vector-valued or nested entries are not supported yet. Future versions ...

would be better?

BJMCox: That reads nicely and makes sense to me 😁

stalbrec: I changed it :)

cburgard:
Results of discussion: 
- make explicit that the order of the categories matters. a category list is a set, which is represented by an array. categories given in a different order are not supposed to have a different meaning. these should be treated as set definitions, not as ordered arrays.
- unbinned data: switch from indices to strings in the entries
- binned data: this is acceptable, because the unrolling is identical

stalbrec:
I clarified that categories are defined as unordered sets and adjusted the unbinned dataset

cburgard:
Great, is this okay to merge now? @oschulz 
From my side, this seems okay.

oschulz:
Ok from my side.

PR #109, opened 2026-07-28
Add Diagonal distributions for compact constraint terms

stalbrec:
## Summary

> This is the PR 3/3 in an effort to split up https://github.com/hep-statistics-serialization-standard/hep-statistics-serialization-standard/pull/105 into smaller, independently reviewable pieces. This part covers diagonal distributions for compact description of constraints paired with auxiliary-measurement data.

Auxiliary/constraint terms are typically declared as one `distributions`/`data` pair per nuisance parameter. For models with many nuisance parameters this multiplies boilerplate almost entirely.

~`DiagonalGauss`/`DiagonalPoisson`~ `DiagonalPdf` collapse same-kind constraint terms into one compact struct, paired with a single `unbinned` data set carrying a categorical axis (one category per constrained parameter) alongside the usual continuous axis. This is semantically equivalent to a *Product distribution* of `n` such terms, but declared flat and explicit so engines *may* use a dedicated, vectorized implementation instead of `n` scalar evaluations.

Both distributions are written to be agnostic about which side of the term (the variable `x` vs. its assigned `mean`) ends up bound to an observable (a paired-data column) versus a floating parameter.

#### What changed
- `2.1.2_composite_distributions.md`: adds ~`DiagonalGauss` and `DiagonalPoisson`, two~ `DiagonalPdf`, one new composite distribution~s~ in. A product of `n` independent one-dimensional ~Gaussian/Poisson~ terms, indexed by a categorical variable rather than declared as `n` separate distributions.
- `2.4_likelihoods.md`: Adds a non-normative "Auxiliary measurements and global observables" section to  document the pattern this PR enables: collecting many same-kind constraint terms (e.g. all unit-Gaussian nuisance-parameter constraints) into a single distribution/data pair instead of one pair per constraint.


docs/chapters/2.4_likelihoods.md
```diff
@@ -43,3 +43,52 @@ The components of a likelihood struct are:
...
+The mechanisms described above, can already acommodate such terms in several ways: as ordinary entry of `distributions` paired with a single-entry data set, as inline numeric value in `data`, or via `aux_distributions` (with the observed value implicit in the definition of the distribution).
+
+For models with many constraints, listing one pair per constraint term can becom lenghty. When the constraint terms are independent and of the same kind (e.g. unit Gaussians or Poisson terms) they can instead be collected into a single pair: an `unbinned` data set with a categorical axis enumerating the constrained parameters, holding one entry per auxiliary measurement, paired with a [DiagonalGauss distribution](#sec:diag-gaussian-distribution){reference-type="ref" reference="sec:diag-gaussian-distribution"} (or [DiagonalPoisson distribution](#sec:diag-poisson-distribution){reference-type="ref" reference="sec:diag-poisson-distribution"}) that evaluates one constraint term per entry:
```
BJMCox: *become lengthy

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,121 @@ The distribution can be specified either via the non-normalized density function
...
+
+
+#### DiagonalGauss distribution {#sec:diag-gaussian-distribution label="sec:diag-gaussian-distribution}
+
+
+The DiagonalGauss distribution is a product of independent, one-dimensional Gaussian distributions over $n$ named dimensions, compactly indexed by a categorical variable rather than declared as $n$ separate distributions:
+
+$$
+\begin{aligned}
+\text{DiagonalGaussPdf}(\vec{x}) &= \prod_{i=1}^{n} \text{GaussianPdf}(x_i; \mu_i, \sigma)
```
BJMCox: What is the reasoning for mu being (potentially) different per entry, but sigma remaining fixed for all? Both scale and location shifting from standard normal can be implemented in vectorised form. It is fine if this is a physics / HEP standard thing of course, given that this is proposed for being used for nuisance parameters I presume this is the case?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,121 @@ The distribution can be specified either via the non-normalized density function
...
+\end{aligned}
+$$
+
+where dimension $i$ is the one whose entry in `dimensions` has `category` equal to the corresponding value of `category_index`, $x_i$ is the variable named by `x` for that dimension, and $\mu_i$ is its assigned `mean`. It is semantically equivalent to a [Product distribution](#sec:product-distribution){reference-type="ref" reference="sec:product-distribution"} of $n$ one-dimensional Gaussians, but declared flat and explicit such that engines [may]{.smallcaps} use a dedicated, vectorized implementation.
```
BJMCox, line 138: This makes sense to make explicit, but it does somewhat beg the question why we do not have a general constructor for vectorised distributions (unless I am missing it). 

Of course Gaussian and Poisson are rather special in that they are the standard distributions used for constraints, but what if one wants a different distribution? Then one must go through all of the machinery to use the product distribution! This is not so much, but a generic convenience constructor could be possible I think (out of scope for this PR of course, but worth bearing in mind in my opinion).

docs/chapters/2.1.2_composite_distributions.md
cburgard: HEP models typically define gaussian constraints to have a standardized width of $\sigma=1$. However, you are of course right that it would be a straight-forward generalization to also allow per-category values of $\sigma$.
It might even be that the syntax that @stalbrec has proposed can already accomodate that. @stalbrec could you comment on that?

docs/chapters/2.1.2_composite_distributions.md
cburgard: General vectorised distributions are a very nice idea, but require some extra thought, because that also plays into the idea of having vector-valued parameters. I suggest you create an issue to discuss the need for this - and if you have a suggestion how you think that could be done in a way that is syntactically elegant and statistically rigorous at the same time, please feel free to propose something!

Besides that, I think that even without generalized vectorized distributions, having diagonal gauss and poisson terms are a good addition, precisely because they are so common and some engines might elect to implement these, but not the generalized vectorized distributions. Would you agree?

BJMCox: I absolutely agree. I noted this as it is a conceptually obvious extension - of course in reality it is not so simple (a big issue of course being, as you mentioned, vector handling)

docs/chapters/2.1.2_composite_distributions.md
stalbrec: We could accommodate this into the individual `dimensions` entries and have a  `default_sigma` (optional - defaults to `1.0`) as fallback in the place of the current `sigma`, e.g. like this:

```
"distributions":
[
{
  "name":"gauss_constraints",
  "type":"diagonal_gauss_dist",
  "category_index":"np",
  "x":"gauss_constraint_x",
  "default_sigma":1.5,
  "dimensions":[
    {"category":"nom_np_1","mean":"np_1","sigma":"np_1_sigma"},
    {"category":"nom_np_2","mean":"np_2"},
    {"category":"nom_np_3","mean":"np_3","sigma":31.622},
]
}
],

```

the `default_sigma`'s motivation would be purely to condense the HS3 files.

cburgard: I like that! @BJMCox what do you think?

BJMCox: Looks good to me 😃

cburgard: two thoughts: do we want to allow `default_sigma` to be a record (name of a parameter), or do we force this to be a literal?
and: do we perhaps also want to explicitly support a `default_mean` as well to put both on the same footing?

BJMCox: I would be in favour of allowing it to be a parameter if possible (unless there is a reason not to) - I would agree that given a `default_sigma` a `default_mean` / `default_mu` (I think mu is used in the base Gaussian distribution here) makes a lot of sense for parity.

stalbrec: I updated both `diagonal_gauss_dist` and `diagonal_poisson_dist` to have per-dimension mean (and sigma) with top-level fallbacks. 
Want to have another look at it @BJMCox @cburgard ? :)

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,129 @@ The distribution can be specified either via the non-normalized density function
...
+                {"name":"gaussian_constraint_x","min":-5, "max":5}
+            ]
+        }
+        ...
+    ]
+}
+```
+
+
+#### DiagonalPoisson distribution {#sec:diag-poisson-distribution label="sec:diag-poisson-distribution}
```
cburgard: I think there's a spare or missing double quote here.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,129 @@ The distribution can be specified either via the non-normalized density function
     -   `type`: `log_density_function_dist` 
     -   `function`: The density function $f$ 
 
+
+
+#### DiagonalGauss distribution {#sec:diag-gaussian-distribution label="sec:diag-gaussian-distribution}
```
cburgard: I think there's a spare or missing double quote here.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,129 @@ The distribution can be specified either via the non-normalized density function
...
+The typical usage, shown below, pairs `category_index` and `x` with the two axes of a data set holding auxilliary-measurement data for the constraint, giving one entry per dimension.
+
+```json title="Example: DiagonalGauss distribution"
+{
+    "distributions":[
+        ...
+            {
+                "name":"gauss_constraints",
+                "type":"diagonal_gauss_dist",
+                "category_index":"np",
```
cburgard: I'm a bit unhappy with the category index being called "np" in this example. People could read this as "the category index is a NP", which is not the point here. Rather, the category index is the index that identifies which NP which dimension belongs to. Hence, I would advocate for renaming the category index from "np" to "np_index" in this example to avoid potential confusion on the side of the reader.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,129 @@ The distribution can be specified either via the non-normalized density function
...
+The typical usage, shown below, pairs `category_index` and `x` with the two axes of a data set holding auxiliary-measurement data for the constraint, giving one entry per dimension.
+
+```json title="Example: DiagonalPoisson distribution"
+{
+    "distributions":[
+        ...
+            {
+                "name":"poisson_constraint",
+                "type":"diagonal_poisson_dist",
+                "category_index":"nom",
```
cburgard: I don't understand why the category index is called "nom" here. Rather, I would advocate for trying to keep this consistent with the DiagonalGauss above and potentially call it "np_index".

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,129 @@ The distribution can be specified either via the non-normalized density function
...
+                "x":"poisson_constraint_x",
+                "default_mean":"np_poisson",
+                "dimensions":[
+                    {"category":"nom_np_1","mean":"np_1"},
```
cburgard: This is completely valid, but one could consider creating a n example that is closer to the common use case. Typically, at least for HistFactory-Like models, the mean of the Poisson is not directly the NP. rather, the mean of the NP is given by some formulaic expression that is a product of two numbers of the following type:
`nominal_sumw2 * np_1`
which would probably be given as a `generic` function or a `product` function.

That way, the NPs can have nominal values around 1, with expected uncertainties given as `sqrt(1/nominal_sumw2)`, which is typically what one would expect. 

This is very specific and should not affect the way the primitive is defined, but perhaps giving this as an example would help people recognize this (common) pattern more easily.

docs/chapters/2.1.2_composite_distributions.md
BJMCox: On my reading what is there looks reasonable 😁

cburgard:
Discussion outcome: instead introduce general diagonal_dist with a structured name/value mapping for parameters

```
of=dist_type, parameters=[{name:a,value:b}]
```

stalbrec:
I unified `diagonal_poisson_dist` and `diagonal_gauss_dist` into `diagonal_dist` with key `of` and `(default_)parameter` struct(s).
Let me know if you thought of it this way @oschulz (also @BJMCox , @cburgard )

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,118 @@ The distribution can be specified either via the non-normalized density function
...
+                ]
+            }
+        ...
+    ],
+    "data":[
+        ...
+        {
+            "name":"nom_np",
+            "type":"unbinned",
+            "entries":[[0,0], [1,0], [2,0]],
```
cburgard: I think we decided that the entries should be strings for the categorical axis, no?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -122,6 +122,118 @@ The distribution can be specified either via the non-normalized density function
...
+            "type":"product",
+            "factors":[10.0, "np_poisson"]
+        }
+    ],
+    "data":[
+        ...
+        {
+            "name":"nom_np",
+            "type":"unbinned",
+            "entries":[[0,1], [1,1]],
```
cburgard: again, I think the categorical entries should be strings

docs/chapters/2.1.2_composite_distributions.md
stalbrec: Thanks for spotting this! I changed it.

docs/chapters/2.1.2_composite_distributions.md
stalbrec: Thanks again!

oschulz:
I'm not quite sure I understand the role of `category_index`. `diagonal_dist` essentially build a product distribution, right? So wouldn't we have multiple independent variates, instead of a "selection" of one variate among many?

cburgard:
> select

I would agree with that assessment, but I don't know where you take the word 'select' from...

oschulz:
> but I don't know where you take the word 'select' from...

Doesn't "categorical variable" imply that there is a one-out-of-many selection?

cburgard:
> > but I don't know where you take the word 'select' from...
> 
> Doesn't "categorical variable" imply that there is a one-out-of-many selection?

My expectation would be that a "draw" from this distribution would give you a categorical dataset, where each category is populated with the result of the "draw" from one of the element distributions. Is this not what the language of this PR communicates to you?

stalbrec:
> > > but I don't know where you take the word 'select' from...
> > 
> > 
> > Doesn't "categorical variable" imply that there is a one-out-of-many selection?
> 
> My expectation would be that a "draw" from this distribution would give you a categorical dataset, where each category is populated with the result of the "draw" from one of the element distributions. Is this not what the language of this PR communicates to you?

@oschulz does this clear up `categorical_index` for you?

oschulz:
Sorry, I had overlooked  @cburgard's reply.

> @oschulz does this clear up categorical_index for you?

I would rather say that the term "categorical" doesn't fit in here. I would argue that a categorical distribution is a distribution where each draw from that distribution results in a random category label. Here, we have a list or set (are they implicitly ordered or not?) of distributions, each associated with a label. Each draw from the diagonal draws once from each of the member distributions. The variate of the diagonal is then a list (does the order matter) or label/variate pairs. Is that combined variate meant to emulate a dictionary or is it an ordered list?

I had originally thought that `diagonal_dist` was meant to be a diagonal product of distributions of the same type but with different parameters and that the variate would be a vector. But the `diagonal_dist` described here, with `"of":"gaussian_dist"`, behaves quite differently from our `multivariate_normal_dist` with a diagonal covariance.

In any case, I doubt that the term "categorical" conveys the right intuition here.

stalbrec:
Thanks @oschulz! I went back and forth on this and I'm not 100% sure I grasped all aspects of this, but I think for me it boils down to this:

- I see that with your definition of a categorical distribution, which is also the usual one, "categorical" implies that the categories follow some distribution. Thinking about this, I noticed that `categorical_dist` in PR 119 (#119) has the wrong name, since there the category must not be sampled. (I'll raise this there)
- For `diagonal_dist` it is not stated otherwise either, so I think `category_index` is underspecified in terms of whether or how the category can be sampled. The options I see:
    - Option A: the category is not sampled; there is a fixed number of entries per category (optionally adding to the spec: either one N for all or N_i for category i, defaulting to 1?)
    - Option B: the category can be sampled, but only if the sub pdfs are extended; the number of entries for category i is then drawn from a Poisson with the expected count of that sub pdf, same as the extended `named_joint_dist` in PR 119

  **I would propose to go with option A and rename `category_index` to `dimension_index`**, since it then only says which entry in `dimensions` a data entry belongs to. With option B we could rename it to `categorical_variate` instead.
- `dimensions` is matched to the data by `category` name, not by position, so as I meant it the order doesn't matter and it's used like a dict. **Should we add that `category` values must be unique?**
- Assuming one entry per category and `of: gaussian_dist`, I think `diagonal_dist` is identical to an MVN with diagonal covariance. The only difference I see is that the MVN has one variate per dimension, while `diagonal_dist` has the observables (c, x), encoded as label value pairs in the unbinned dataset with a categorical axis.

PR #111, opened 2026-07-29
Introduce authoring guidelines for new HS3 primitives

cburgard:
## Motivation

As HS³ continues to grow, an increasing number of new primitive types are being proposed. While the existing specification defines the *syntax* and *semantics* of the primitives themselves, it currently provides little guidance on **how new primitives should be specified**.

This PR introduces a new appendix that establishes authoring guidelines for future primitive specifications.

The goal is not to prescribe a rigid document template, but rather to define a quality standard for new additions to the specification.

## Main ideas

The appendix emphasizes that HS³ is fundamentally a mathematical interchange format. Primitive specifications should therefore describe the mathematical semantics of a model component independently of any particular software framework whenever possible.

At the same time, the specification should remain useful for several different audiences:

* software engineers implementing importers, exporters or evaluators in new inference frameworks,
* domain experts mapping existing software concepts onto HS3,
* statisticians inspecting serialized models to understand their mathematical structure.

The amount of documentation required for a primitive should be proportional to its complexity. A simple closed-form function may require little more than its mathematical definition and JSON syntax, whereas more sophisticated primitives—such as wrappers around external software packages or algorithmic model representations—may require substantially more explanation in order to permit independent implementations.

## Guiding principle

The appendix proposes a practical criterion for evaluating new primitive specifications:

> A primitive specification should contain sufficient mathematical, algorithmic and semantic information that independent implementations can reproduce equivalent behaviour without requiring additional clarification from the primitive authors.

This criterion intentionally focuses on interoperability rather than implementation details.

## Scope

This PR is purely editorial.

It introduces guidance for future contributors but does **not** modify the syntax or semantics of any existing HS3 primitive.

The appendix is intended to improve consistency across future extensions while preserving the concise style already used throughout the specification for simple mathematical primitives.

### DIFF
--- a/docs/chapters/3.2_authoring_primitives.md
+++ b/docs/chapters/3.2_authoring_primitives.md
@@ -1,5 +1,5 @@
 ---
-bibliography: ./hs3.bib
+bibliography: hs3.bib
 ---
 
 
--- a/docs/chapters/3.2_authoring_primitives.md
+++ b/docs/chapters/3.2_authoring_primitives.md
@@ -14,7 +14,7 @@ Primitive specifications define the mathematical semantics of new HS³ primitive
 Primitive specifications should be understandable to three audiences simultaneously:
 
 - **Software implementers**, who should be able to implement import, export, and evaluation directly from the specification.
-- **Domain experts**, who should be recognize how concepts from existing software correspond to the HS³ representation.
+- **Domain experts**, who should recognize how concepts from existing software correspond to the HS³ representation.
 - **Statisticians**, who should be able to infer the mathematical properties of the primitive directly from its specification and serialized representation.
 
 The mathematical definition is normative. Implementation-specific terminology may be used where it improves understanding, but should complement rather than replace the mathematical semantics.
--- a/docs/chapters/3.2_authoring_primitives.md
+++ b/docs/chapters/3.2_authoring_primitives.md
@@ -46,7 +46,7 @@ Every primitive specification shall define:
 
 Primitive specifications should define the normative semantics of a primitive independently of any particular implementation. Whenever practical, this should be done by specifying the mathematical object represented by the primitive. Where no sufficiently precise mathematical characterization exists, or where the algorithm itself defines the abstraction, a normative algorithmic specification may be used instead.
 
-The amount of explanatory material should be proportional to the complexity of the primitive. Closed-form mathematical primitives generally require little more than their definition and serialization, whereas implementation-backed primitives shall include whatever additional explanation is necessary for independent implementations. Informative mappings to existing software are encouraged but never replace the mathematical definition.
+The amount of explanatory material should be proportional to the complexity of the primitive. Simple primitives may require little more than their mathematical definition and serialization. More complex primitives should include whatever additional explanation is necessary to enable independent conforming implementations. Informative mappings to existing software are encouraged but never replace the normative definition.
 
 ### Serialization
 
--- a/docs/chapters/3.2_authoring_primitives.md
+++ b/docs/chapters/3.2_authoring_primitives.md
@@ -3,7 +3,7 @@ bibliography: ./hs3.bib
 ---
 
 
-## Authoring Requirements for HS³ Primitive Specifications
+## Authoring Requirements for HS³ Primitive Specifications {#sec:primitive_authoring_requirements} 
 
 This appendix is **normative**.
 
--- a/docs/chapters/3.2_authoring_primitives.md
+++ b/docs/chapters/3.2_authoring_primitives.md
@@ -88,6 +88,6 @@ Normative text defines behaviour. Informative text may provide motivation, examp
 
 As a practical guideline, contributors should continuously ask:
 
-> Could an independent implementation of this primitive be written correctly from this specification alone, together with the documentation of any explicitly referenced external libraries?
+> Could an independent, faithful implementation of this primitive be written without ambiguity from this specification alone, together with the documentation of any explicitly referenced external libraries?
 
 If the answer is no, the specification is likely missing semantic or algorithmic information.
--- a/docs/chapters/3.2_authoring_primitives.md
+++ b/docs/chapters/3.2_authoring_primitives.md
@@ -3,18 +3,18 @@ bibliography: hs3.bib
 ---
 
 
-## Authoring Requirements for HS³ Primitive Specifications {#sec:primitive_authoring_requirements} 
+## Authoring Requirements for HS$^3$ Primitive Specifications {#sec:primitive_authoring_requirements} 
 
 This appendix is **normative**.
 
-Primitive specifications define the mathematical semantics of new HS³ primitives. Their purpose is to enable independent implementations to interpret serialized models consistently and without ambiguity.
+Primitive specifications define the mathematical semantics of new HS$^3$ primitives. Their purpose is to enable independent implementations to interpret serialized models consistently and without ambiguity.
 
 ### Intended audience
 
 Primitive specifications [should]{.smallcaps} be understandable to three audiences simultaneously:
 
 - **Software implementers**, who should be able to implement import, export, and evaluation directly from the specification.
-- **Domain experts**, who should recognize how concepts from existing software correspond to the HS³ representation.
+- **Domain experts**, who should recognize how concepts from existing software correspond to the HS$^3$ representation.
 - **Statisticians**, who should be able to infer the mathematical properties of the primitive directly from its specification and serialized representation.
 
 The mathematical definition is normative. Implementation-specific terminology may be used where it improves understanding, but [should]{.smallcaps} complement rather than replace the mathematical semantics.
@@ -25,15 +25,15 @@ A primitive specifies **what** a model computes, not **how** a particular implem
 
 Specifications shall therefore define primitives in mathematical terms whenever practical. Implementation-specific concepts may be included only where they are necessary to faithfully represent existing software.
 
-A primitive that introduces functionality which cannot already be represented using existing HS³ primitives will generally be appropriate, provided its semantics are well defined.
+A primitive that introduces functionality which cannot already be represented using existing HS$^3$ primitives will generally be appropriate, provided its semantics are well defined.
 
 A primitive that only specializes an existing abstraction [should]{.smallcaps} provide clear additional value, such as a substantially more compact representation of a common use case, access to significantly more efficient or numerically robust evaluation strategies, or improved interoperability through the explicit representation of a widely recognized abstraction.
 
-### Relationship to the HS³ language
+### Relationship to the HS$^3$ language
 
-Primitive specifications extend the vocabulary of HS³ but do not redefine the language itself.
+Primitive specifications extend the vocabulary of HS$^3$ but do not redefine the language itself.
 
-Primitive specifications may introduce new functions, distributions, or regular members of any other existing top-level HS³ category (see Section @sec:toplevel). They shall not introduce new language constructs. In particular, a primitive shall not change its semantic role based on configuration or context, create or transform other serialized objects, or introduce control-flow, parser directives, or other metaprogramming facilities. Such proposals constitute extensions of the HS³ language and require separate specification and review.
+Primitive specifications may introduce new functions, distributions, or regular members of any other existing top-level HS$^3$ category (see Section @sec:toplevel). They shall not introduce new language constructs. In particular, a primitive shall not change its semantic role based on configuration or context, create or transform other serialized objects, or introduce control-flow, parser directives, or other metaprogramming facilities. Such proposals constitute extensions of the HS$^3$ language and require separate specification and review.
 
 ### Required specification
 
@@ -68,7 +68,7 @@ Information that has no effect on mathematical evaluation shall be stored in `mi
 
 Where a primitive represents functionality already established in an existing framework or scientific domain, its semantics [should]{.smallcaps} remain recognizable to domain experts while preserving implementation-independent mathematical meaning.
 
-Primitive specifications [should]{.smallcaps} minimize the conceptual distance between established domain abstractions and their HS³ representation. Generalization is appropriate where it improves interoperability or removes accidental implementation details. It [should]{.smallcaps} be avoided where it obscures established abstractions or complicates import, export, or implementation without providing corresponding semantic benefit.
+Primitive specifications [should]{.smallcaps} minimize the conceptual distance between established domain abstractions and their HS$^3$ representation. Generalization is appropriate where it improves interoperability or removes accidental implementation details. It [should]{.smallcaps} be avoided where it obscures established abstractions or complicates import, export, or implementation without providing corresponding semantic benefit.
 
 ### Alternative implementations
 
--- a/docs/chapters/3.2_authoring_primitives.md
+++ b/docs/chapters/3.2_authoring_primitives.md
@@ -33,9 +33,7 @@ A primitive that only specializes an existing abstraction should provide clear a
 
 Primitive specifications extend the vocabulary of HS³ but do not redefine the language itself.
 
-A primitive shall belong to an existing HS³ semantic category and inherit the evaluation and composition semantics of that category.
-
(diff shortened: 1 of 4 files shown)


cburgard:
I've written down a few of the design principles I think we should put forward as requirements for new primitives to be added. I think in the long run, our project would benefit massively from having clear guidelines on how we want extensions of our standard to look.

I'd be interested in some community review of this - please feel free to suggest even complete rewrites if you feel this is necessary. I think getting these principles right will massively impact long-term sustainability of HS3.

To get started, if @kratsg @oschulz @guitargeek @stalbrec @BJMCox could take a look at this, I'd be thrilled!

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+
+The mathematical definition is normative. Implementation-specific terminology may be used where it improves understanding, but should complement rather than replace the mathematical semantics.
+
+### Design principles
+
+A primitive specifies **what** a model computes, not **how** a particular implementation computes it.
+
+Specifications shall therefore define primitives in mathematical terms whenever practical. Implementation-specific concepts may be included only where they are necessary to faithfully represent existing software.
+
+A primitive should introduce only the semantic concept that is genuinely new. Prediction models, statistical interpretations, normalization conventions, and composition rules should remain separate whenever they can already be expressed using existing HS³ primitives.
```
BJMCox: I would say potentially that "already be expressed using existing HS³ primitives" is rather broad - for example the DiagGauss and DiagPoisson (sp.) PR could be expressed using existing primitives. I think that primitives can/should be introduced if it is either (a) a new concept, (b) an existing concept for which a (very?) common special case can is hard/clunky to implement or (c) there is a case where implementation through the existing path is numerically unstable and a stable implementation exists but is not fully general

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Relationship to the HS³ language
+
+Primitive specifications extend the vocabulary of HS³ but do not redefine the language itself.
```
BJMCox, line 34: This is rather strange wording in my opinion - I think that what is meant is that new specifications are meant to be purely additive (i.e. non breaking)?

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+
+A primitive shall belong to an existing HS³ semantic category and inherit the evaluation and composition semantics of that category.
```
BJMCox: It might then be worth clarifying these semantic categories and their evaluation / composition categories normatively

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+Every primitive specification shall define:
+
+- its mathematical semantics, or an equivalent algorithmic definition;
+- the meaning of every serialized field, including the relationship between serialized objects and the mathematical quantities they represent;
+- any normative default values;
+- sufficient examples to remove ambiguity.
+
+Whenever practical, primitives should be defined by closed-form mathematical expressions. Algorithmic definitions should be used only where no compact mathematical formulation exists.
+
+The amount of explanatory material should be proportional to the complexity of the primitive. Closed-form mathematical primitives generally require little more than their definition and serialization, whereas implementation-backed primitives shall include whatever additional explanation is necessary for independent implementations. Informative mappings to existing software are encouraged but never replace the mathematical definition.
```
BJMCox: I would disagree with "Closed-form mathematical primitives generally require little more than their definition and serialization". For example, the PDF of the multivariate normal is rather simple to write down, but there are many many ways to evaluate it depending on the structure of the covariance matrix, and the obvious way of inverting the covariance matrix as in the closed form is basically the worst in terms of numerical stability and computational cost. I would say that this is in implementation detail, however things like this are probably worth a note in the definitions, as otherwise implementations may diverge and give differing results due to numerical stability?

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Serialization
```
BJMCox, line 51: This sections seems a bit at odds with itself. Shall is normative language meaning at all times with no exceptions, but we then carve out exceptions and preferences? I think we may wish to loosen the language in the first sentence here to "should" rather than "shall" - it seems that it is somewhat unavoidable for now that some specific things will be serialised.

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+
+A primitive specification is complete if independent implementations can reproduce equivalent mathematical behaviour from the specification alone, together with the documentation of any explicitly referenced external libraries.
+
+Normative text defines behaviour. Informative text may provide motivation, examples, implementation guidance, historical context, or mappings to existing software, but does not contribute to the formal definition of the primitive.
+
+### Editorial Guideline
+
+As a practical guideline, contributors should continuously ask:
+
+> Could an independent implementation of this primitive be written correctly from this specification alone, together with the documentation of any explicitly referenced external libraries?
```
BJMCox: I think this is a good test, however mathematical equivalence != numerical equivalence. I think we want to converge on one of these (mathematical is likely better as choosing a normative algorithmic implementation for everything is tiresome and prone to issues such as different paradigms being more amenable to different implementations, CPU compared to GPU being the obvious example).

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+
+Primitive specifications should minimize the conceptual distance between established domain abstractions and their HS³ representation. Generalization is appropriate where it improves interoperability or removes accidental implementation details. It should be avoided where it obscures established abstractions or complicates import, export, or implementation without providing corresponding semantic benefit.
+
+### Alternative implementations
+
+An enum-like field may select between alternative implementations only if the alternatives represent the same mathematical abstraction, operate on substantially the same serialized state, and differ only in evaluation strategy.
+
+Alternatives requiring different parameterizations, serialized structures, mathematical assumptions, or validity conditions shall be represented by separate primitives.
+
+The meaning of every alternative shall be defined normatively. Unsupported alternatives shall either produce an explicit error or an explicitly reported substitution; silent substitution is not permitted.
```
BJMCox: I would favour this being some kind of switch (like -Werror=...) with a documented normative default

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Alternative implementations
+
+An enum-like field may select between alternative implementations only if the alternatives represent the same mathematical abstraction, operate on substantially the same serialized state, and differ only in evaluation strategy.
+
+Alternatives requiring different parameterizations, serialized structures, mathematical assumptions, or validity conditions shall be represented by separate primitives.
+
+The meaning of every alternative shall be defined normatively. Unsupported alternatives shall either produce an explicit error or an explicitly reported substitution; silent substitution is not permitted. 
+
+The ability to define a reasonably direct transformation between two alternatives is often a good indication that they represent different implementations of the same abstraction rather than distinct primitives.
```
BJMCox: I note that the converse statement is very much untrue, hence my comment on line 49. This is generally a good test but the direct transformation is typically very much in the mathematics, which can obscure some algorithmic choices (choices within a given implementation, such as a choice of solver for a subproblem). I think these choices could be encoded in the misc information, but then we risk the misc information becoming bloated and carrying many variables each relevant to only one implementation.

BJMCox:
Aside from a few minor nits with the wording, looks good to me 😄

docs/chapters/3.2_authoring_primitives.md
cburgard: You are absolutely right. I have revised this section to reflect these insights.

docs/chapters/3.2_authoring_primitives.md
cburgard: It's a bit more than that. What I was aiming for was what is explicated by the following paragraph (see below).

docs/chapters/3.2_authoring_primitives.md
cburgard: I agree that the wording was unclear. I have rewritten this paragraph to be more precise in what I was trying to say.

docs/chapters/3.2_authoring_primitives.md
cburgard: Good point. I don't think the distinction I wanted to make is actually "mathematical vs. algorithmic", but rather "normative semantics vs. implementation strategy".

The intent is that a primitive specification should prescribe what is computed, not how it is computed. In most cases this is naturally expressed as a mathematical definition. However, there are cases where the algorithm itself is the mathematical definition of the object (for example functions defined by an iterative construction). In those cases a normative algorithmic specification is entirely appropriate.

docs/chapters/3.2_authoring_primitives.md
cburgard: I think you are right about this. Upon closer inspection, the second paragraph conveys little information. I removed it.

docs/chapters/3.2_authoring_primitives.md
cburgard: Good point. I agree that implementations should not silently ignore the requested method, and that users should be able to detect or reject substitutions.

I think the more fundamental point that was missing from the text is that conforming implementations are not required to support every method defined by a primitive. I have made that explicit now.

Rather than prescribing a particular diagnostics interface (e.g. a warning switch), I'd instead make it normative that an implementation which does not support the requested method must either reject the object or explicitly report the use of a semantically valid fallback. The mechanism by which this is exposed to the caller (warnings, exceptions, status codes, command-line options, etc.) should remain implementation-defined.

docs/chapters/3.2_authoring_primitives.md
cburgard: That makes sense to me, I reworded the paragraph a bit also.

docs/chapters/3.2_authoring_primitives.md
BJMCox: This makes sense to me - as long as it is traceable where and why the implementation elected to change a method (I would think mostly due to not being implemented) it should be okay.

docs/chapters/3.2_authoring_primitives.md
cburgard: Good point. I don't think "correctly" is the right word here, since it can be read as requiring a particular numerical implementation rather than agreement on the intended semantics.

More generally, I think the distinction throughout this document should be between the normative semantics of a primitive and the implementation strategy used to realize it. In most cases the semantics are best expressed mathematically, but in some cases they may themselves be algorithmic. What matters is that independent implementations converge on the same intended behaviour, not that they use the same numerical procedure.

I reprhased this as

> Could an independent, faithful implementation of this primitive be written without ambiguity from this specification alone, together with the documentation of any explicitly referenced external libraries?

cburgard:
> Aside from a few minor nits with the wording, looks good to me 😄

Thank you for the insightful review. I have addressed all your comments. Please resolve the conversations where you are happy with the changes, and reply to the others.

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,93 @@
...
+This appendix is **normative**.
+
+Primitive specifications define the mathematical semantics of new HS³ primitives. Their purpose is to enable independent implementations to interpret serialized models consistently and without ambiguity.
+
+## Intended audience
+
+Primitive specifications should be understandable to three audiences simultaneously:
+
+- **Software implementers**, who should be able to implement import, export, and evaluation directly from the specification.
+- **Domain experts**, who should be recognize how concepts from existing software correspond to the HS³ representation.
```
stalbrec: The sentence is off; I guess the "be" is leftover.
```suggestion
- **Domain experts**, who should recognize how concepts from existing software correspond to the HS³ representation.
```

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,93 @@
...
+- **Statisticians**, who should be able to infer the mathematical properties of the primitive directly from its specification and serialized representation.
+
+The mathematical definition is normative. Implementation-specific terminology may be used where it improves understanding, but should complement rather than replace the mathematical semantics.
+
+### Design principles
```
stalbrec: I suppose `Design principles` should not be nested under `Intended audience`, correct?

```suggestion
## Design principles
```

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Authoring Requirements for HS³ Primitive Specifications {#sec:primitive_authoring_requirements} 
+
+This appendix is **normative**.
+
+Primitive specifications define the mathematical semantics of new HS³ primitives. Their purpose is to enable independent implementations to interpret serialized models consistently and without ambiguity.
+
+## Intended audience
+
+Primitive specifications should be understandable to three audiences simultaneously:
```
stalbrec: Do we want RFC keywords to be formatted the same way as in the remaining docs? (i.e. `[should]{.smallcaps}`, etc.)

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,93 @@
...
+---
+
+
+## Authoring Requirements for HS³ Primitive Specifications {#sec:primitive_authoring_requirements} 
+
+This appendix is **normative**.
+
+Primitive specifications define the mathematical semantics of new HS³ primitives. Their purpose is to enable independent implementations to interpret serialized models consistently and without ambiguity.
+
+## Intended audience
```
stalbrec: I suppose this should be a `<h3>`, right?
```suggestion
### Intended audience
```

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,93 @@
+---
+bibliography: hs3.bib
+---
+
+
+## Authoring Requirements for HS³ Primitive Specifications {#sec:primitive_authoring_requirements}
```
stalbrec: I think for the remaining official documentation that gets digested by pandoc and mkdocs we still use latex for the superscript (i.e. `HS$^3$`).
I saw that #110 changed it to unicode for the README, but thats just because of githubs md-renderer, right?

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Alternative implementations
+
+An enum-like field may select between alternative methods when those methods operate on the same serialized state and the meaning of all other fields remains unchanged. The selected method may alter how that state is interpreted or evaluated, but shall not determine which disjoint set of fields is applicable.
+
+If different enum values require substantially different configuration data, parameterizations, or validity conditions, they should generally be represented as separate primitives rather than as variants of a single primitive.
+
+Implementations need support only a subset of the methods defined by a primitive specification. An implementation that does not support the requested method shall not silently substitute another method. It shall either reject the object or explicitly report the use of a semantically valid fallback.
```
stalbrec: Is this missing a "to"?

```suggestion
Implementations need to support only a subset of the methods defined by a primitive specification. An implementation that does not support the requested method shall not silently substitute another method. It shall either reject the object or explicitly report the use of a semantically valid fallback.
```

stalbrec:
Sorry for the delay in getting to this. Overall i like this appendix. Just a few minor comments on style/gramma (see inline).

docs/chapters/3.2_authoring_primitives.md
cburgard: I have applied this to "should" at all the places I found appropriate, please check.

docs/chapters/3.2_authoring_primitives.md
cburgard: yes, fixed.

docs/chapters/3.2_authoring_primitives.md
```diff
@@ -0,0 +1,93 @@
+---
+bibliography: hs3.bib
+---
+
+
+## Authoring Requirements for HS$^3$ Primitive Specifications {#sec:primitive_authoring_requirements} 
+
+This appendix is **normative**.
```
cburgard: remove this

cburgard:
Reference these on the README

PR #116, opened 2026-08-28
Add `analytic_convolution_dist` and canonical resolution models

Phmonski:
# Add FFT and decay convolution distributions

## Summary

This PR specifies two convolution distributions with related terminology but
different input contracts:

- `fft_convolution_dist` references general source and kernel distributions;
- `decay_convolution_dist` directly contains a restricted exponential decay
  profile and an inline `gaussian`, `delta`, or `mixture` kernel.

An FFT convolution has the outer components `name`, `type`, `x`, `source`, and
`kernel`. A decay convolution has the flat outer components `name`, `type`,
`x`, `envelope`, `terms`, and `kernel`. The decay representation introduces no
helper source distribution or function node.

## Mathematical contract

For a continuous convolution variate $x$, source density $S$, and kernel
$K$, both convolution types represent

$$
q(x)=\int_{-\infty}^{\infty}S(u)K(x-u)\,du,
\qquad
p(x)=\frac{q(x)}{\mathcal M}.
$$

The inner integral spans the full real line. The outer normalization
$\mathcal M$ follows the current HS<sup>3</sup> domain. Both convolution
types are non-extended; event yields use the existing extended-distribution
mechanism.

The resulting density must be finite and non-negative throughout its domain
and must not be identically zero.

## FFT convolution

An FFT convolution has:

- `name`: custom unique string;
- `type`: `fft_convolution_dist`;
- `x`: convolution variate;
- `source`: source-distribution reference; and
- `kernel`: kernel-distribution reference.

The source and kernel must be non-extended distributions and must both use
`x` as their convolution variate.

```json
{
  "name": "smeared_signal",
  "type": "fft_convolution_dist",
  "x": "mass",
  "source": "unsmeared_signal",
  "kernel": "detector_response"
}
```

## Decay convolution

A decay convolution has:

- `name`: custom unique string;
- `type`: `decay_convolution_dist`;
- `x`: convolution variate;
- `envelope`: inline exponential-envelope object;
- `terms`: non-empty array of inline decay terms; and
- `kernel`: inline kernel object.

There is no separate source node. The envelope and terms directly define the
unconvolved profile

$$
S(u)=E(u)A(u),
\qquad
A(u)=\sum_{i=1}^{n}c_iB_i(u).
$$

### Exponential envelope

`envelope` contains a positive `scale` and a `support` equal to `positive`,
`negative`, or `both`. It has no `type`: every envelope in this distribution
is exponential. For \(\tau=\text{scale}\), the alternatives are

$$
E_+(u;\tau)=\mathbf 1_{u\geq0}e^{-u/\tau},\qquad
E_-(u;\tau)=\mathbf 1_{u\leq0}e^{u/\tau},\qquad
E_{\pm}(u;\tau)=e^{-|u|/\tau}.
$$

### Additive terms

Every entry of `terms` contains exactly:

- `coefficient`: a number or real-valued parameter or function reference
  representing $c_i$; and
- `basis`: a supported inline basis object representing $B_i$.

The conversion from serialized terms to the additive sum is direct. For each
entry, multiply the value of `coefficient` by the function represented by
`basis`; then add all of those products:

$$
\left[
  \{\text{coefficient}:c_1,\text{basis}:B_1\},\ldots,
  \{\text{coefficient}:c_n,\text{basis}:B_n\}
\right]
\quad\longmapsto\quad
A(u)=c_1B_1(u)+\cdots+c_nB_n(u).
$$

A `constant` basis represents $B_i(u)=1$ and has no other component. A
`sin`, `cos`, `sinh`, or `cosh` basis contains `rate` and represents
$B_i(u)=g(\text{rate}\,u)$. Derived rates are represented by ordinary
function references.

The scale, rates, and coefficients must be independent of `x`. Individual
coefficients may be negative, but the complete unconvolved and convolved
densities must be finite, non-negative, and nonzero.

### Simple decay with a delta kernel

```json
{
  "name": "two_sided_decay",
  "type": "decay_convolution_dist",
  "x": "decay_time",
  "envelope": {
    "scale": "lifetime",
    "support": "both"
  },
  "terms": [
    {
      "coefficient": 1.0,
      "basis": {"type": "constant"}
    }
  ],
  "kernel": {"type": "delta"}
}
```

### Four-term decay with a Gaussian kernel

```json
{
  "name": "four_term_decay",
  "type": "decay_convolution_dist",
  "x": "decay_time",
  "envelope": {
    "scale": "lifetime",
    "support": "positive"
  },
  "terms": [
    {
      "coefficient": "cosh_coefficient",
      "basis": {"type": "cosh", "rate": "half_delta_gamma"}
    },
    {
      "coefficient": "sinh_coefficient",
      "basis": {"type": "sinh", "rate": "half_delta_gamma"}
    },
    {
      "coefficient": "cos_coefficient",
      "basis": {"type": "cos", "rate": "delta_mass"}
    },
    {
      "coefficient": "sin_coefficient",
      "basis": {"type": "sin", "rate": "delta_mass"}
    }
  ],
  "kernel": {
    "type": "gaussian",
    "mean": "time_bias",
    "sigma": "time_resolution_width"
  }
}
```

### Inline kernels

The kernel catalog contains:

- `{"type": "gaussian", "mean": ..., "sigma": ...}`, with positive
  `sigma`;
- `{"type": "delta"}`; and
- `{"type": "mixture", "summands": [...], "coefficients": [...]}`.

A mixture recursively contains inline kernels. For $n$ summands, it contains
either $n$ coefficients that sum to one or $n-1$ coefficients, with the
last coefficient defined as $1-\sum_{i=1}^{n-1}c_i$. Every explicit and
implicit coefficient must be non-negative. Kernel parameters must be
independent of `x`.

```json
{
  "name": "mixture_smeared_decay",
  "type": "decay_convolution_dist",
  "x": "decay_time",
  "envelope": {
    "scale": "lifetime",
    "support": "positive"
  },
  "terms": [
    {
      "coefficient": 1.0,
      "basis": {"type": "constant"}
    }
  ],
  "kernel": {
    "type": "mixture",
    "summands": [
      {"type": "gaussian", "mean": 0.0, "sigma": "core_width"},
      {"type": "gaussian", "mean": 0.0, "sigma": "tail_width"}
    ],
    "coefficients": ["core_fraction"]
  }
}
```


docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -37,32 +37,215 @@ $$
...
+  "name": "four_term_decay",
+  "type": "decay_convolution_dist",
+  "x": "decay_time",
+  "envelope": {
+    "scale": "lifetime",
+    "support": "positive"
+  },
+  "terms": [
+    {
+      "coefficient": "cosh_coefficient",
```
cburgard: Do you ever define what this key does?

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -37,32 +37,215 @@ $$
...
+-   `terms`: non-empty array of inline decay-term objects
+-   `kernel`: inline kernel object satisfying the kernel catalog below
+
+The unconvolved decay profile is
+
+$$
+S(u)=E(u)\sum_{i=1}^{n}c_iB_i(u),
+$$
+
+where `envelope` defines $E$, and each entry of `terms` defines one pair $(c_i,B_i)$. The `envelope` object has a positive `scale` and `support` equal to `positive`, `negative`, or `both`. Since every supported envelope is exponential, it has no `type` component.
```
cburgard, line 90: I'm not convinced that "positive", "negative" and "both" is the best way to write this.
Just reading "both" is not very expressive - people would need to look up that this means "both parts of the axis".
I would personally suggest we use slightly different keys here, namely: "-0", "-0+" and "0+", such as to indicate the parts of the axis that constitute the support. This also leaves open the possibility to introduce the keys "+" and "-" to deliberately exclude 0.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -37,32 +37,215 @@ $$
...
+    "scale": "lifetime",
+    "support": "positive"
+  },
+  "terms": [
+    {
+      "coefficient": "cosh_coefficient",
+      "basis": {"type": "cosh", "rate": "half_delta_gamma"}
+    },
+    {
+      "coefficient": "sinh_coefficient",
```
cburgard, line 202: I think the names of the coefficients are chosen a bit awkwardly here. From the example, it is not clear that these are arbitrary strings / parameter / function names. Perhaps you can vary them a bit between the terms, such as using "1.", "parameter_A" and "parametrization_func" such as to span the realm of possibilities a bit better and make clear what they can be used for.

docs/chapters/2.1.2_composite_distributions.md
cburgard, line 90: The inline objects for `envelope` and `terms` are woefully underdefined here. I would spend at least one paragraph each explaining in detail how these can be defined and what the allowed keys are, with a bullet-`components`-list shown for each.

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -37,32 +37,215 @@ $$
...
+
+#### Decay convolution distribution {#dist:decay-convolution label="dist:decay-convolution"}
+
+A decay convolution distribution describes the convolution of a exponential decay profile with a restricted inline kernel. It has the following components:
+
+-   `name`: custom unique string
+-   `type`: `decay_convolution_dist`
+-   `x`: name of the convolution variate $x$
+-   `envelope`: inline exponential-envelope object
+-   `terms`: non-empty array of inline decay-term objects
```
cburgard, line 81: Are you sure `terms` is the best word here?
`terms` is very general and could mean anything. From what I understand, what you are building here is essentially a linear combination, so perhaps `linear_combination` would be better, or, in-keeping with other `sum`-type syntax in HS3, `summands` (or, if you want to stick with `terms`, potentially `summand_terms`)?

cburgard (commented):
Thanks for the next round of refactoring. 
I have a another round of rather nitpicky comments.

PR #119, opened 2026-09-01
Introduce indexed distributions, function definitions and indexed data sets.

stalbrec:
## Summary

> This is PR 2/3 in an effort to split up #105 into smaller, independently reviewable pieces. This part covers indexed data sets as well as indexed distributions and functions.

Building on the categorical axes from #108, this adds the constructs that combine data sets and distributions across them. Such a combination can already be expressed as a likelihood, i.e. a paired list $\{(m_i, x_i)\}$. However a likelihood cannot be sampled from, since sampling is defined on a distribution. A paired list also has no place to record the index itself, neither its name nor its category labels, so what is one indexed object in the engines comes back as $N$ unrelated pairs. Indexed data sets and indexed distributions carry the same content as a single data set and a single distribution, with the index explicit.

Whether the index is an observable or a model parameter is not declared: it follows from how the model is paired with data in a `likelihood`. One definition therefore covers both simultaneous multi-region fits and discrete profiling over alternative shape hypotheses. `indexed_dist` maps onto `RooSimultaneous`, now that most of the functionality of `RooMultiPdf` has been merged into it (root-project/root#23142), and the `indexed` function onto [`RooMultiReal`](https://root.cern.ch/doc/v640/RooMultiReal_8h_source.html).

> **Note**: `combine` uses `RooMultiPdf` for discrete profiling over shape hypothesis and implements additional penalty terms tied to the different choices of shapes (dependent on the shapes complexity). These could be serialized as an `indexed` function used by a distribution in `aux_distributions`. Once an implementation for this is ready in RooFit/Combine we could add non-normative prose on this to the standard.

#### What changed
- `2.3_data.md`: adds the `indexed` data set which comprises of sub-data sets of any type, each assigned to a `coordinates` tuple in the space spanned by categorical `axes` as defined in the *Axis Specifications* section.
- `2.1.2_composite_distributions.md`: adds `indexed_dist`, $\text{PDF}(m_c(\theta), x)$, with the sub-distribution selected by the categorical variable named by `index` and one entry per category in `dimensions`. A sub-distribution may itself be an `indexed_dist` on a different index, in any order, so that nesting matches a data set spanned by several categorical axes.
- `2.2_functions.md`: adds the `indexed` function, structurally identical to `indexed_dist`

### DIFF
--- a/docs/chapters/2.1.2_composite_distributions.md
+++ b/docs/chapters/2.1.2_composite_distributions.md
@@ -7,7 +7,7 @@ bibliography: hs3.bib
 This section contains composite distributions in the sense that they refer to other distribution which they combine or modify in some way. 
 
 
-#### Mixture distribution
+#### Mixture distribution {#sec:mixture-distribution label="sec:mixture-distribution"}
 
 The mixture distribution, sometimes called addition of distributions, is a (weighted) sum of distributions $f_i(\vec{x})$, depending on the same variable(s) $\vec{x}$: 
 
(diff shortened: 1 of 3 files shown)


docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -61,7 +61,81 @@ If all coefficients are given, their sum is used as the expected number of event
...
--   `extended`: boolean denoting whether this is an extended model
+-   `extended`: boolean denoting whether this is an extended model  
+
+
+#### Indexed distribution {#sec:indexed-distribution label="sec:indexed-distribution"}
+
+The indexed distribution combines several sub-distributions, selected by the value $c$ of the categorical variable named by `index`:
+
+$$
+\begin{aligned}
```
BJMCox, line 74: As it stands right now, this integrates to `c` ($\sum_c \int \text{PDF}(m_c, x)\,dx = K$), which conflicts with the our general statement that distributions are normalised. This is a conditional density in $x$ given $c$, not a normalised (joint) density. 

This could be fixed by normalising per category conditioning on `c` (I think?)

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -61,7 +61,81 @@ If all coefficients are given, their sum is used as the expected number of event
...
+\text{IndexedPdf}(x, c) &= \text{PDF}(m_{c}(\theta), x)
+\end{aligned}
+$$
+
+where $m_c$ is the sub-distribution assigned to the category $c$.
+
+An indexed distribution is a struct defined by the following components:
+
+- `name`: custom string
```
BJMCox: This is required to be unique, right? I think we make that explicit elsewhere

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -61,7 +61,81 @@ If all coefficients are given, their sum is used as the expected number of event
...
+- `type`: `indexed_dist`
+- `index`: name of the categorical variable selecting the sub-distribution. It [must]{.smallcaps} refer to a categorical axis of the data this distribution is paired with, or to a parameter with a categorical domain
```
BJMCox: We should also note the precedence order here (if categorical axis named as per `index` that is used, otherwise the parameter named as `index` is used (badly written but you get the point 😄 ) )

docs/chapters/2.1.2_composite_distributions.md
```diff
@@ -61,7 +61,81 @@ If all coefficients are given, their sum is used as the expected number of event
...
+- `dimensions`: array of structs, each assigning one category value to the distribution used for it:
+    - `category`: string, one of the values of the categorical variable named by `index`
+    - `dist`: name of the sub-distribution used for this value of `category`
+
+This standard does not distinguish the two cases: if `index` is bound to a categorical axis of the paired data, evaluation is dispatched per data entry, which is the typical construction for simultaneous fits of several measurement regions; if `index` is a model parameter, the single sub-distribution matching its current value is evaluated. Both follow from the definition above.
```
BJMCox: We have not introduced the two cases yet - consider reordering things so that the cases are introduced before we speak about them.

docs/chapters/2.2_functions.md
```diff
@@ -49,6 +49,35 @@ $$
...
+### Indexed Function
+A value $f_c$ selected by the value $c$ of a categorical variable.  
+
+$$
+\begin{aligned} \text{Indexed}(c) &= f_c \end{aligned}
+$$
+
+-   `name`: custom unique string
+-   `type`: `indexed`
+-   `index`: name of the categorical variable selecting the value
```
BJMCox: We should explicitly constrain what `index` can refer to, as we did in 2.1.2

docs/chapters/2.1.2_composite_distributions.md
stalbrec: You are right! 
I added some text and changed it to $\text{IndexedPdf}(x, c) = w_c \cdot \text{PDF}(m_{c}(\theta), x)$. With $w_c = \nu_c / \sum_{c'} \nu_{c'}$ with $\nu_c$ being the expected number   for sub-dist assigend to $c$, if all sub-dists are extended; and $w_c=1$ otherwise, i.e. conditioning on $c$. This mirrors what `RooSimultaneous` does (If i understand correctly)

What do you think about this?

docs/chapters/2.1.2_composite_distributions.md
stalbrec: fixed

docs/chapters/2.2_functions.md
stalbrec: fixed

docs/chapters/2.1.2_composite_distributions.md
stalbrec: Should the precedence be part of the spec? If both would exist; the engine would map it to what it needs for a certain task, right?

docs/chapters/2.1.2_composite_distributions.md
stalbrec: I have put this to the motivational/descriptional part in the beginning and softened the phrasing.

docs/chapters/2.1.2_composite_distributions.md
stalbrec: I'd say this is related/the same issue to the notion of the *analysis context* in #101

stalbrec:
@cburgard @Phmonski @BJMCox @oschulz @szabo137 @guitargeek : following @cburgard suggestion above, I renamed the key of the `categorical` function from `categorical_parameter` to `categorical_variable`.

Are we in agreement? If there are no further objections, I'd consider this ready to merge.

stalbrec:
Before this goes in, I did one more careful read and pushed two small fixes (sorry for the extra round):

- `dimensions` entries in `named_joint_dist`, `categorical_dist` and the `categorical` function now have to have unique `category` values, matching the existing rule for the `combined` data set.
- Categories without an entry were "undefined", but the authoring guidelines require evaluation to be fully determined. This is now an error instead: for `named_joint_dist`, the paired data must not contain entries in such a category; for `categorical_dist` and the `categorical` function, the categorical variable must not take such a value.

There are two more points where I'd like your opinion:

1. In the discussion on `diagonal_dist` with @oschulz in #109, I noticed that `categorical_dist` has a misleading name. By the usual definition, a categorical distribution is a distribution over categories, so each draw gives a random category. In our `categorical_dist` the category is not the variate but a parameter: it selects one of several sub distributions, and it must not be sampled. I would propose to rename it, for example to `switch_dist` or `multi_dist`. **Do you have a preference?**

2. It's currently not specified what happens in `named_joint_dist` when only some sub distributions are extended. I would propose that each extended sub distribution keeps its own extended term, which also keeps it compatible with how `RooSimultaneous` builds the likelihood per channel. **Does that sound reasonable?**

PR #124, opened 2026-09-08
Histfactory Specfication Sync

stalbrec:
In `histfactory_dist`:
- Drops the singular `parameter`/`constraint`; modifiers use `parameters` and `constraints` (parallel arrays, `null` allowed for unconstrained parameters). Entries in `constraints` name a distribution in the top-level `distributions` list instead of a constraint *type* string.
- Drops `data.vals` of `shapesys`; the per-bin uncertainties live in the referenced constraint distributions.
- States the nominal-value convention (0 for `histosys`/`normsys`, 1 otherwise) and that `histosys`/`normsys` constraints [should] be unit Gaussians.
- `staterror`: defines the per-bin relative uncertainty from sample `contents`/`errors` and that its constraints [should] be a `barlow_beeston_lite_poisson_constraint_dist`.
- `axes` with the binned-data axis rules of §2.3 (`name`, `min`, `max` required; exactly one of `nbins` or `edges`) 
- marks sample `data.errors` as optional.

**Known gaps vs. ROOT master**: ROOT writes `data.vals` for `shapesys`, writes no `constraints` for `staterror` (rebuilds them on import), reads a channel-level `staterror` struct (`relThreshold`, `constraint_type`) that the spec does not define, and does not emit `barlow_beeston_lite_poisson_constraint_dist` yet.

### DIFF
--- a/docs/parts/histfactory.md
+++ b/docs/parts/histfactory.md
@@ -126,13 +126,12 @@ The components of a HistFactory distribution are:
     
 -   `name`: custom unique string 
 -   `type`: `histfactory_dist` 
--   `axes`: array of structs representing the axes. If given each struct     needs to have the component `name`. Further,     ([optional]{.smallcaps}) components are `max`, `min` and `nbins`,     or, alternatively, `edges`. The definition of the axes follows the     format for binned data (see Section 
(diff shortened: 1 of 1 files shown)


stalbrec:
@cburgard @Phmonski Would you be able to have a look at this?

